PR #12, opened 2018-02-11
Fix _WD_Window

Danp2:


Danp2:
- Fixed: _WD_Window implementation of Maximize, Minimize, Fullscreen, & Screenshot
- Removed: Normal option from _WD_Window
- Added: Rect option to _WD_Window

Issue #44, opened 2020-10-27
_WD_Screenshot Base64


seadoggie01:
It looks like nothing is returned when using _WD_Screenshot with $nOutputType = 3... I think that $sResult was supposed to be set to $sResponse under Case 3

Danp2:
Took a quick look and I think you're correct. Want to make a PR with the fix?

PR #45, opened 2020-10-27
Chore: Update function headers

seadoggie01:
~~(No code changes)~~
Added descriptions
Removed extra (empty) lines
Added _WD_Screenshot return values

N.B.: Diff for these changes doesn't appear to be working


seadoggie01:
Added commit for fixing _WD_ScreenShot (Issue #44). I can split these if you don't want them both in the same pull request.

PR #46, opened 2020-10-29
_WD_ElementOptionSelect - SetErrors

seadoggie01:
Added Return SetError to capture and re-throw error.
Updated function header to include all possible returned errors.

### DIFF
--- a/wd_helper.au3
+++ b/wd_helper.au3
@@ -852,7 +852,10 @@ EndFunc
 ;                  $sStartElement       - [optional] Element ID of element to use as starting point
 ; Return values .: None
 ;                  @ERROR       - $_WD_ERROR_Success
+;                  				- $_WD_ERROR_Exception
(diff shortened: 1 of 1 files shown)


wd_helper.au3
```diff
@@ -852,7 +852,10 @@ EndFunc
 ;                  $sStartElement       - [optional] Element ID of element to use as starting point
 ; Return values .: None
 ;                  @ERROR       - $_WD_ERROR_Success
+;                  				- $_WD_ERROR_Exception
```
Danp2, line 855: These should be aligned with the prior one

Danp2 (commented):
Thanks for the submission. Looks good other than the alignment issue

seadoggie01:
I think that this appears only because of how tabs appear on GitHub... it makes tabs appear equal to 8 spaces. Your other functions use four tabs after the initial spaces as well. They appear mis-aligned as well :)

PR #47, opened 2020-11-06
Change (__WD_CloseDriver): Call ProcessWaitClose

Danp2:
Call ProcessWaitClose after each attempt to close a process with ProcessClose


Danp2:
@seadoggie01 Can you test this PR to see if it solves the issue we discussed on the forum?

PR #49, opened 2021-03-12
Update ConsoleWrites

seadoggie01:
_WD_Startup was using __WD_ConsoleWrite("_WDStartup... instead of $sFuncName
Updated for accuracy and consistency


Danp2:
Thanks for catching that!

PR #56, opened 2021-04-06
_WD_Attach() - Refactoring

mlipok:
Moved parameters validation to the top of function.
Error checking first.


mlipok:
Do not need to be suplemented , will stick with your current version.

PR #58, opened 2021-04-06
_WD_FrameEnter - fix for {"id":null}

mlipok:
Fixed NULL usage and added Remarks

### DIFF
--- a/wd_helper.au3
+++ b/wd_helper.au3
@@ -499,7 +499,7 @@ Func _WD_IsWindowTop($sSession)
 EndFunc   ;==>_WD_IsWindowTop
 
 ; #FUNCTION# ====================================================================================================================
-; Name ..........: _WD_FrameEnter
+; Name ..........: _WD_FrameEnter 
 ; Description ...: This will enter the specified frame for subsequent WebDriver operations.
 ; Syntax ........: _WD_FrameEnter($sSession, $vIdentifier)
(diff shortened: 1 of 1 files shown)


Danp2 (commented):
Thanks for catching my error with null. Regarding your "remarks" section, I'm unsure if it helps or just adds to the confusion.

wd_helper.au3
```diff
@@ -509,8 +509,8 @@ EndFunc   ;==>_WD_IsWindowTop
 ;                  @ERROR       - $_WD_ERROR_Success
 ;                  				- $_WD_ERROR_Exception
 ; Author ........: Decibel
-; Modified ......: 2018-04-27
-; Remarks .......:
+; Modified ......: mLipok
+; Remarks .......: For nested frames use this function as many time as many nesting level is in analyzed website
```
mlipok, line 513: `You can drill-down into nested frames by calling this function repeatedly with the correct parameters`
will be better

Danp2: @mlipok Pretty sure that you just submit an additional commit to the same branch (https://github.com/mlipok/WebDriver/tree/patch-4)

mlipok: I hope I do exactly what you mean, if not please provide me a hint.

Danp2 (approved):

PR #59, opened 2021-04-07
#include "" instead #include <>

mlipok:
This two following include files
```
#include <JSON.au3>
#include <WinHttp.au3>
```

are not Standard AutoIt UDF so should be used "" instead <>
```
#include "JSON.au3"
#include "WinHttp.au3"
```

according to helpfile
https://www.autoitscript.com/autoit3/files/beta/autoit/docs/keywords/include.htm

```
If "..." is used, the filename is taken to be relative to the current script.
If <...> is used the filename is taken to be relative to include library directory (usually C:\Program Files\AutoIt3\Include). The include library contains many pre-written user-functions for you to use!
```

### DIFF
--- a/wd_core.au3
+++ b/wd_core.au3
@@ -1,7 +1,7 @@
 #include-once
 #include <WinAPIProc.au3>
-#include <JSON.au3> ; https://www.autoitscript.com/forum/topic/148114-a-non-strict-json-udf-jsmn
-#include <WinHttp.au3> ; https://www.autoitscript.com/forum/topic/84133-winhttp-functions/
+#include "JSON.au3" ; https://www.autoitscript.com/forum/topic/148114-a-non-strict-json-udf-jsmn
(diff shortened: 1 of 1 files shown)


Danp2:
Not sure that I agree with your assessment. Also according to the help file --

image

So, either option is valid. It's just the search order that is affected.

Danp2:
IMO, this isn't a situation where one method is right or wrong. However, I can see where it could be advantageous to have the search order reversed so that the script directory is searched first.

mlipok:
Yeah.
When you build project with specific include file version then you put them into the same directory as main script.
and then it is prefered to use them when you use `#include ""`

For the sam reason all standard UDF use  `#include ""` internally like I show to you in  `WinAPIProc.au3`
This is required also for "AutoIt portable" and for use with #AuitoItWrapper directive like in the following example:
```
#AutoIt3Wrapper_AutoIt3Dir="z:\AutoItPortable\AutoIt_3_3_15_1"
```

PR #65, opened 2021-04-24
Unnecessary Code

seadoggie01:
No need to set/save/restore the debug level to none inside this function ;)


Danp2 (approved):
This was leftover from when _WD_IsLatestRelease used __WD_Get ;-)

PR #67, opened 2021-04-26
updated links to: Discussion Thread on Autoit Forums

mlipok:


Danp2 (changes requested):
The links don't wrap correctly when viewed on Github. It appears to me that this is caused by a LF vs CR/LF issue. Please fix.

Danp2 (changes requested):
First entry has incorrect link ;-)

Danp2 (approved):

PR #69, opened 2021-04-27
Fix (_WD_ElementActionEx): Properly terminate JSON string

Danp2:


Danp2:
@seadoggie01 Please test and report back.

PR #75, opened 2021-06-18
_WD_GetSession($sSession) 	added #forceref $sSession, $sFuncName

mlipok:
For Au3Check compatibility


Danp2:
Please explain why this is needed. What settings cause AU3Check to flag these?

PR #76, opened 2021-06-18
wd_helper.au3 ++ _WD_GetBrowserVersion() ++ _WD_GetWebDriverVersion()

mlipok:
These feature have been separated into separate functions to make it possible for the user to check it by themself.

```AutoIt
MsgBox(0, "", _WD_GetBrowserVersion('chrome') & @CRLF & _
		"@error = " & @error & @CRLF & _
		"@extended = " & @extended & @CRLF & _
		"")
MsgBox(0, "", _WD_GetWebDriverVersion(@ScriptDir,'chromedriver.exe') & @CRLF & _
		"@error = " & @error & @CRLF & _
		"@extended = " & @extended & @CRLF & _
		"")

```

### DIFF
--- a/wd_helper.au3
+++ b/wd_helper.au3
@@ -1263,11 +1263,9 @@ EndFunc   ;==>_WD_IsLatestRelease
 ; ===============================================================================================================================
 Func _WD_UpdateDriver($sBrowser, $sInstallDir = Default, $bFlag64 = Default, $bForce = Default)
 	Local Const $sFuncName = "_WD_UpdateDriver"
-	Local $iErr = $_WD_ERROR_Success, $sEXE, $sDriverEXE, $sPath, $sBrowserVersion, $sCmd, $iPID, $bResult = False
-	Local $sOutput, $sDriverVersion, $sVersionShort, $sDriverLatest, $sURLNewDriver
+	Local $iErr = $_WD_ERROR_Success, $sDriverEXE, $sBrowserVersion, $bResult = False
+	Local $sDriverVersion, $sVersionShort, $sDriverLatest, $sURLNewDriver
 	Local $sReturned, $sTempFile, $hFile, $oShell, $FilesInZip, $sResult, $iStartPos, $iConversion
-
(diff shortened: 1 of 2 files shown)


wd_helper.au3
```diff
@@ -35,7 +35,7 @@
 #cs
 	- Jonathan Bennett and the AutoIt Team
 	- Thorsten Willert, author of FF.au3, which I've used as a model
-	- Michał Lipok (@mLipok) for all his feedback / suggestions
+	- Michał Lipok (@mLipok) for all his feedback / suggestions / contribution
```
Danp2, line 38: You are thanking yourself? LOL

Danp2:
I'll have to take a closer review of your proposed changes, but a few things on my initial observation --

- There are changes to other functions that are unrelated to the described PR. 
- Since the proposed new functions are being "ripped" from _WD_UpdateDriver, listing yourself as the author is a stretch ;-)

wd_helper.au3
mlipok: Now I have tweaked a little for a less extensive description.
But I wonder why I do not see here my recent change (from about few minutes ago)

mlipok (commented):
ah I see here is my small change (above this post)

mlipok:
> * There are changes to other functions that are unrelated to the described PR.


```autoit
	Local $sData = InetRead($sURL, $iOptions)

```
this part was modified because InetRead states:
`Failure: | "" (empty string) and sets the @error flag to non-zero.
`
Which mean that you can not be sure which value will be set to @error

mlipok:
> * Since the proposed new functions are being "ripped" from _WD_UpdateDriver, listing yourself as the author is a stretch ;-)

fixed

Danp2:
> this part was modified because helpfile for InetRead says:
> `Failure: | "" (empty string) and sets the @error flag to non-zero. `
> Which mean that you can not be sure which value will be set to @error

That's fine, but IMO it should have been a separate PR because it doesn't have anything to do with the _WD_GetBrowserVersion() and _WD_GetWebDriverVersion(), which is what this PR is concerning.

mlipok:
I hope that now all is fine, because I still learn how github works.

mlipok:
Something went wrong with "Resolve conflict" will check it.

PR #83, opened 2021-07-25
wd_core.au3 - $_WD_DefaultTimeout - not as Const

mlipok:
User should be able to adjust this default value.
I can imagine that someone has a very poor connection and is doing a project, e.g. got a freelancer contract, and now has to adjust to a slower connection.


Danp2 (changes requested):
The user can currently override the default timeout by providing the new value for each relevant function call. However, I can see where this would have value as well.

This update will require a few additional changes --

- Move the line to the correct region (global constant vs global variable)
- Update _WD_Option to allow the user to retrieve / set the default timeout

mlipok:
give me one week... I'm on vacation.

mlipok:
I had time for creativity, please review

Danp2 (approved):

PR #87, opened 2021-09-03
TIDY code

mlipok:
just using TIDY befor making next changes/proposals


Danp2:
Thanks, but I'm thinking that it would be easier for me to push this from my end instead of attempting to merge your changes back into my dev environment.

PR #89, opened 2021-09-12
Add: _WD_CheckContext

Danp2:


mlipok (commented):
$lReconnect
Should be named $bReconnect

Danp2:
> 
> 
> $lReconnect
> Should be named $bReconnect

Old habits are hard to break. 😀

PR #92, opened 2021-09-22
first official release of wd_capabilities.au3

mlipok:
This UDF was disscused here:
https://www.autoitscript.com/forum/topic/206576-wd_capabilitiesau3-support-topic-early-beta-version-work-in-progress/

Demo /example script was added as separate PR:
https://github.com/Danp2/WebDriver/pull/93

But then I found out how to add them as single PR.


Danp2:
@mlipok Why the separate demo for wd_capabilities? I'd prefer to have everything in wd_demo. Did you consider that?

mlipok:
Will take a look in comming few days.

mlipok:
I can move to wd_demo.au3
But for now, I'd like some feedback.
Is the current demo good in terms of how this UDF works?

mlipok:
Is this following a way which you mean ?
```

Local $aDemoSuite[][2] = _
		[ _
		["DemoTimeouts", False], _
		["DemoNavigation", True], _
		["DemoElements", False], _
		["DemoScript", False], _
		["DemoCookies", False], _
		["DemoAlerts", False], _
		["DemoFrames", False], _
		["DemoActions", False], _
		["DemoDownload", False], _
		["DemoWindows", False], _
		["DemoUpload", False], _
		["DemoAdvanceCapabilities", False] _
		]


```

Danp2:
The existing demo is there to show the user some basic functionality; it isn't intended to demonstrate every feature that exists. 

>  Is the current demo good in terms of how this UDF works?

I'll have to review your current demo to understand it's functionality. 

> Is this following a way which you mean ?

Actually, I was just thinking that the "Setup" functions for each browser could be updated so that the DesiredCapabilities strings would be generated instead of hardcoded.

mlipok:
> Actually, I was just thinking that the "Setup" functions for each browser could be updated so that the DesiredCapabilities strings would be generated instead of hardcoded.

This was my next proposal, so here we agree.

Danp2:
> So when to share/post the advanced example, on the wiki ?

wiki, [here](https://www.autoitscript.com/forum/topic/201537-webdriver-example-scripts-collection/), or another forum thread

mlipok:
Please take a look for this:
https://github.com/Danp2/WebDriver/pull/97

this is semi step to add changes your propose in relation to Capabilities demo.

mlipok:
> I'll have to review your current demo to understand it's functionality.

`wd_capabilities_demo.au3` will be shared to the community on AutoIt forum like you propose, but firstly `wd_capabilities.au3` should became a part of this UDF, so we need goes ahead to commit this PR to the main UDF source code.

> Actually, I was just thinking that the "Setup" functions for each browser could be updated so that the DesiredCapabilities strings would be generated instead of hardcoded.

Done.

mlipok:
btw.
Do README.md should be suplemented ?

wd_helper.au3
```diff
@@ -1,7 +1,11 @@
 #include-once
+; standard UDF's
 #include <File.au3> ; Needed For _WD_UpdateDriver
 #include <InetConstants.au3>
+
+; WebDriver related UDF's
 #include "wd_core.au3"
+#include "wd_capabilities.au3"
```
Danp2, line 8: I would expect `wd_capabilities.au3` to be included here.

wd_demo.au3
```diff
@@ -398,7 +400,11 @@ Func SetupGecko()
 	_WD_Option('DriverParams', '--log trace')
 	_WD_Option('Port', 4444)
 
-	Local $sDesiredCapabilities = '{"capabilities": {"alwaysMatch": {"browserName": "firefox", "acceptInsecureCerts":true}}}'
+	_WD_CapabilitiesStartup()
+	_WD_CapabilitiesAdd('alwaysMatch', 'firefox')
+	_WD_CapabilitiesAdd('w3c', True)
+	_WD_CapabilitiesAdd('excludeSwitches', 'enable-automation')
```
Danp2, line 406: Not sure that this is even valid for Firefox. Shouldn't it be `acceptInsecureCerts` like in the line you are replacing? Also, I suggest that you somehow leave the original desiredCapabilities string (probably just a comment) so that the user can know what you intend to generate with the _WD_Capabilities* calls.

Danp2 (changes requested):

Danp2:
> Do README.md should be suplemented ?

Yes, with a section of _WD_Capabilities functions

wd_helper.au3
Danp2: Corrected above comment to reflect that wd_capabilities.au3 should *not* be included in wd_helper. wd_core.au3 is needed because it contains functions called by the "helper" functtions, but this isn't true about wd_capabilities.au3

wd_capabilities.au3
```diff
@@ -1,5 +1,11 @@
+#include-once
+
+; standard UDF's
 #include <MsgBoxConstants.au3>
+
+; WebDriver related UDF's
 #include "wd_helper.au3" ; https://github.com/Danp2/WebDriver
```
Danp2, line 7: Don't think this is needed

Danp2 (changes requested):

wd_capabilities.au3
mlipok: I take a look in next few hours.

wd_helper.au3
mlipok: So user of WD***.au3 should intentionally include helper and capabilities as a separate files ?

wd_demo.au3
mlipok: sure, done

wd_capabilities.au3
mlipok: done

wd_helper.au3
Danp2: Yes. wd_core.au3 should function independent of the other "add on" modules. If the user needs more functionality, then they have the option to #include one or more files.

Danp2 (approved):

mlipok:
> > Do README.md should be suplemented ?
> 
> Yes, with a section of _WD_Capabilities functions

Done, 

ps.
I hope that all things you point me was corrected

Danp2 (approved):

wd_capabilities.au3
Danp2: This line still needs removed from wd_capabilities AFAICS

Danp2 (changes requested):

mlipok:
just forget to commit, but now should be fine

Danp2 (approved):

mlipok:
with Edge:
`"session not created: This version of MSEdgeDriver only supports MSEdge version 93\nCurrent browser version is 96.0.1054.41 with binary path C:\\Program Files (x86)\\Microsoft\\Edge\\Application\\msedge.exe"`

Danp2:
I encountered the same thing a few days ago. If you autoupdate, then you are hiding the issue from the script developer. What do you think about adding a checkbox to the GUI so that the update is only performed if the user has checked it?

mlipok:
> What do you think about adding a checkbox to the GUI so that the update is only performed if the user has checked it?

Yes this is good idea, but in separate PR as this is not related to wd_capabilities.au3

mlipok:
In a future I also plan to add demo for _WD_Sleep()

PR #94, opened 2021-10-10
__WD_Sleep .. SetError($_WD_ERROR_UserAbort)

mlipok:
__WS_Sleep() should  set $_WD_ERROR_UserAbort in case of error instead $_WD_ERROR_Timeout


mlipok:
now I think this is all fine 
As so far I do not plan other changes in this PR

mlipok:
in 
Refactoring __WD_Sleep() usage in functions STEP2
I ony use TIDY nothing more.

mlipok:
> * Why do you use $v_Parameter for custom _My_Sleep function? Why not $iDelay?
Fixed

mlipok:
So if $iDelay=100 why there was:
`__WD_Sleep(1000)
`
It's a bit weird when I say wait a little second $iDelay=100 and then the program waits 10 times longer.

btw. 
FunctionHeader says
`$iDelay              - [optional] Milliseconds to wait before checking status`

looking in the code and I see that status is checked multiple time in a While loop.

Danp2:
>  It's a bit weird when I say wait a little second $iDelay=100 and then the program waits 10 times longer.

It's an optional parameter that allows the user to not start the process of checking for the desired element for a specific length of time. In it's default setting, the $iDelay value equals 0, so there isn't a delay. If the user passes in something like 5000, then there is an initial delay before entering the loop.

The __WD_Sleep(1000) is simply a pause during the looping process. It could be any other value (ie: 250, 500), but I didn't see an issue with a 1 sec delay per loop cycle

>  FunctionHeader says

Yes, the header could be clearer. Could be as simple as adding the word "initially" before "checking status".

> looking in the code and I see that status is checked multiple time in a While loop.

Correct. What issue do you have with that?

mlipok:
> > It's a bit weird when I say wait a little second $iDelay=100 and then the program waits 10 times longer.
> 
> It's an optional parameter that allows the user to not start the process of checking for the desired element for a specific length of time. In it's default setting, the $iDelay value equals 0, so there isn't a delay. If the user passes in something like 5000, then there is an initial delay before entering the loop.
> 
> The __WD_Sleep(1000) is simply a pause during the looping process. It could be any other value (ie: 250, 500), but I didn't see an issue with a 1 sec delay per loop cycle
> 
> > FunctionHeader says
> 
> Yes, the header could be clearer. Could be as simple as adding the word "initially" before "checking status".
> 
> > looking in the code and I see that status is checked multiple time in a While loop.
> 
> Correct. What issue do you have with that?

Will answer in next few days

mlipok:
> > FunctionHeader says
> 
> Yes, the header could be clearer. Could be as simple as adding the word "initially" before "checking status".

done

mlipok:
> > The __WD_Sleep(1000) is simply a pause during the looping process. It could be any other value (ie: 250, 500), but I didn't see an issue with a 1 sec delay per loop cycle
> > > looking in the code and I see that status is checked multiple time in a While loop.
> > 
> > 
> > Correct. What issue do you have with that?
> 
> Working on example .....


`
#include <GuiComboBoxEx.au3>
#include <GUIConstantsEx.au3>
#include <ButtonConstants.au3>
#include <MsgBoxConstants.au3>
#include <WindowsConstants.au3>

#include "wd_helper.au3"
;~ #include "WebDriver-master\wd_helper.au3"

Global $sSession

_Example()
If @error Then ConsoleWrite("! ---> @error=" & @error & "  @extended=" & @extended & _
		" : Error" & @CRLF)

Func _Example()
	SetupChrome()
	_WD_Navigate($sSession, 'https://google.com')
	If @error Then Return SetError(@ScriptLineNumber)

	MsgBox($MB_OK + $MB_TOPMOST + $MB_ICONINFORMATION, "Information #" & @ScriptLineNumber, "Accept the POP UP message in Google Chrome")

	Local $hTimer = TimerInit() ; Begin the timer and store the handle in a variable.
	For $i = 1 To 10
		_Search()
		If @error Then Return SetError(@error)
	Next
	Local $fDiff = TimerDiff($hTimer) ; Find the difference in time from the previous call of TimerInit. The variable we stored the TimerInit handlem is passed as the "handle" to TimerDiff.
	MsgBox($MB_SYSTEMMODAL, "Time Difference", $fDiff / 1000 & ' Seconds')
EndFunc   ;==>_Example

Func _Search()
	_WD_Navigate($sSession, 'https://google.com')
	If @error Then Return SetError(@ScriptLineNumber)

	_WD_LoadWait($sSession)
	If @error Then Return SetError(@ScriptLineNumber)

	Local $sElement_Input = _WD_WaitElement($sSession, $_WD_LOCATOR_ByXPath, '//*[@name="q"]')
	If @error Then Return SetError(@ScriptLineNumber)

	_WD_SetElementValue($sSession, $sElement_Input, 'AutoIt', $_WD_OPTION_Advanced)
	If @error Then Return SetError(@ScriptLineNumber)

	Sleep(10)
	_WD_ElementAction($sSession, $sElement_Input, 'click')

	Local $sElement_Search = _WD_WaitElement($sSession, $_WD_LOCATOR_ByXPath, '//input[@name="btnK"]', Default, Default, $_WD_OPTION_Enabled)
	If @error Then Return SetError(@ScriptLineNumber)

	Do
		Sleep(10)
		_WD_ElementAction($sSession, $sElement_Search, 'click')
	Until Not @error

	Local $sElement_PaginatorNext = _WD_WaitElement($sSession, $_WD_LOCATOR_ByXPath, '//*[@id="pnnext"]')
	If @error Then Return SetError(@ScriptLineNumber)

	_WD_ElementAction($sSession, $sElement_PaginatorNext, 'click')
	If @error Then Return SetError(@ScriptLineNumber)

	_WD_LoadWait($sSession)
	If @error Then Return SetError(@ScriptLineNumber)

EndFunc   ;==>_Search

Func SetupChrome()
	_WD_UpdateDriver('chrome')
	_WD_Option('Driver', @ScriptDir & '\chromedriver.exe')
	_WD_Option('Port', 9515)
	_WD_Option('DriverParams', '--verbose --log-path="' & @ScriptDir & '\chrome.log"')
	_WD_Startup()

	Local $sDesiredCapabilities = '{"capabilities": {"alwaysMatch": {"goog:chromeOptions": {"w3c": true, "excludeSwitches": [ "enable-automation"]}}}}'
	$sSession = _WD_CreateSession($sDesiredCapabilities)
	If @error <> $_WD_ERROR_Success Then Exit
EndFunc   ;==>SetupChrome

`

mlipok:
Ok. So now I have proper explantion which show why I do not want to use:
`__WD_Sleep(1000)`

Check this following example with and without this 1000 ms sleep.
[wd_testing_Sleep.zip](https://github.com/Danp2/WebDriver/files/7584796/wd_testing_Sleep.zip)

mlipok:
> The __WD_Sleep(1000) is simply a pause during the looping process. It could be any other value (ie: 250, 500), but I didn't see an issue with a 1 sec delay per loop cycle

My example shows that after each proper finding there is added 1 sec waiting, where there is no need to wait so long.
I could understand to add:
`__WD_Sleep(10)` which will not overheat CPU, and will keep interaction (script responsivity) thanks to using:
`_WD_Option('Sleep', _MyResponsiveSleepFunction)`

wd_helper.au3
```diff
@@ -330,9 +330,9 @@ Func _WD_WaitElement($sSession, $sStrategy, $sSelector, $iDelay = Default, $iTim
 	If $bCheckNoMatch And $iOptions <> $_WD_OPTION_NoMatch Then
 		$iErr = $_WD_ERROR_InvalidArgue
 	Else
+		__WD_Sleep($iDelay)
```
Danp2, line 333: In a earlier revision, we had added some error checking here

wd_helper.au3
```diff
@@ -723,9 +724,9 @@ Func _WD_LoadWait($sSession, $iDelay = Default, $iTimeout = Default, $sElement =
 	If $iTimeout = Default Then $iTimeout = $_WD_DefaultTimeout
 	If $sElement = Default Then $sElement = ""
 
+	__WD_Sleep($iDelay)
```
Danp2, line 727: Same here re: error checking for user abort

Danp2 (changes requested):

Danp2 (approved):

PR #97, opened 2021-11-27
wd_demo.au3 - Au3Check - compatibility

mlipok:


mlipok:
just added:
_WD_Demo()
moved some declaration to `Global`, some was not strictly defined by `Local` some was not used so I use `#forceref`

wd_demo.au3
```diff
@@ -5,18 +5,19 @@
...
-Local Const $sElementSelector = "//input[@name='q']"
+#Region - Global's declarations
 
-Local $sDesiredCapabilities, $iIndex, $sSession
-Local $nMsg, $bProcess = False
-
-Local $aBrowsers[][2] = _
-		[["Firefox", SetupGecko], _
+Global Const $sElementSelector = "//input[@name='q']"
+Global Const $aBrowsers[][2] = _
```
Danp2, line 11: These globals could be moved to locals of new _WD_Demo function

Danp2 (approved):
See comment re: globals, which isn't a big deal.

wd_demo.au3
mlipok: checking

mlipok: $sElementSelector is used in `DemoElements()` and `DemoActions()` so I would'nt want to move them.
What is your opionion ?

$aBrowsers is used only in `_WD_Demo()` but I want to split them to 2 spearate function and do not exit main look in another way then using $GUI_EVENT_CLOSE.

Danp2: Right... I was referring to the globals below $sElementSelector. 

> $aBrowsers is used only in _WD_Demo() but I want to split them to 2 spearate function and do not exit main loop in another way then using $GUI_EVENT_CLOSE.

Not sure what you mean here. Please explain.

mlipok: For this moment I porpose that you accept this PR as it is.
Then I do my next PR which will show to you exactly what I mean.

Danp2 (commented):
I don't believe that you need to merge the changes from the master branch into your patch branch. Thoughts?

mlipok:
It was just that easier for me.

PR #99, opened 2021-11-27
wd_helper.au3 - __WD_Sleep() @error checking

mlipok:
unification for @error checking with _WD_Sleep()


mlipok:
This is PREVIEW will need some more checking

Danp2:
I don't believe that these changes are necessary / desirable. Please review our prior discussion about $iDelay [here](https://github.com/Danp2/WebDriver/pull/94#issuecomment-940791224).

PR #108, opened 2021-12-12
wd_demo.au3 + DemoSleep + Abort button

mlipok:


Danp2:
@mlipok Thanks for the submission. I think this may need to be retooled a bit --

- I think `_WD_Option("Sleep", _USER_WD_Sleep)` should be moved into the DemoSleep function. Otherwise, this will affect all demo functions and not just DemoSleep. You would save then restore the setting at the end of the function.

- Instead of adding an Abort button to the GUI, couldn't the exisitng Run Demo button be repurposed temporarily? 

- Instead of issuing `Exit` inside _USER_WD_Sleep, shouldn't this event flow back into `RunDemo` and then handled accordingly?

mlipok:
> * I think `_WD_Option("Sleep", _USER_WD_Sleep)` should be moved into the DemoSleep function. Otherwise, this will affect all demo functions and not just DemoSleep. You would save then restore the setting at the end of the function.

I can only check if DemoSleep button is checked and then make the use of this function dependent on this button, but I cannot move `_WD_Option("Sleep", _USER_WD_Sleep)` beyond/outside of "SetUp ...." functions.


> * Instead of adding an Abort button to the GUI, couldn't the exisitng Run Demo button be repurposed temporarily?

You mean the "Run" button when clicked would turn into an "Abort" button when "DemoSleep" was selected, and back int "Run" button when procesing ends ?

I would like to do this differently I would like to stick with Run and with Abort button but Abort should be Disalbed/Enabled in relation to state of "DemoSleep".

I'm waiting for opinions.

> * Instead of issuing `Exit` inside _USER_WD_Sleep, shouldn't this event flow back into `RunDemo` and then handled accordingly?

As so far $GUI_EVENT_CLOSE was doing alomst the same.
Back into 'RunDemo' is action revelant to Abort button.

Do you have other thoughts? What is your intention?

Danp2:
> I can only check if DemoSleep button is checked and then make the use of this function dependent on this button, but I cannot move _WD_Option("Sleep", _USER_WD_Sleep) beyond/outside of "SetUp ...." functions.

I don't understand why this can't be moved. Please explain your position.

> You mean the "Run" button when clicked would turn into an "Abort" button when "DemoSleep" was selected, and back int "Run" button when procesing ends ?

Yes, exactly this. I was also thinking that we may want to disable this button while other demos are running.

> I would like to do this differently I would like to stick with Run and with Abort button but Abort should be Disalbed/Enabled in relation to state of "DemoSleep".

Ok as long as button is disabled when outside of DemoSleep routine

> Do you have other thoughts? What is your intention?

My goal was to demonstrate the proper technique of passing the exit message up the call stack. In your current implementation, the script just abruptly stops without performing the final steps of RunDemo.

mlipok:
> I don't understand why this can't be moved. Please explain your position.

check current commit

mlipok:
> Ok as long as button is disabled when outside of DemoSleep routine

done.

mlipok:
> * Otherwise, this will affect all demo functions and not just DemoSleep

Affect ? Yes. Is it a problem ?
"DemoSleep" demonstrate it specificaly... I mean you can exit long procesing in _WD_WaitElement()
But this is not a problem for using this feature in other demo's function.

Danp2:
> > Ok as long as button is disabled when outside of DemoSleep routine

> done... also other or at least I hope that I done them all.

Nope... not even close to what I stated as a requirement. You are enabling the Abort button when Demo Sleep routine is checked by user. I wanted the button enabled **only** within the DemoSleep function. 🙄

I feel like you are making this more complicated than necessary.  

- I don't think RunDemo needs the extra parameter
- Get rid of `	Local $bDemoSleep = False`
- This should be inside DemoSleep function - `_WD_Option("Sleep", _USER_WD_Sleep)`
- Need to restore prior `_WD_Option("Sleep")` when exiting DemoSleep function

I will need to give some more thought to the use of $__g_bInternalExit variable. My initial thinking was that it could be done without any globals. Just notify the calling routine by setting @error to an appropriate value.

mlipok:
> I wanted the button enabled **only** within the DemoSleep function.

aha.... ok.



> * This should be inside DemoSleep function - `_WD_Option("Sleep", _USER_WD_Sleep)`

Till now my reasoning about how WD UDF works in this specific case: `_WD_Option("Sleep", _USER_WD_Sleep)` 
Was that it should be used before `_WD_Startup()` and before `_WD_CreateSession($sDesiredCapabilities)`

Both: `_WD_Startup()` and `_WD_CreateSession($sDesiredCapabilities)` are called before `DemoSleep`  function is called by `Call($aDemoSuite[$iIndex][0])` thus I thought that this must be used this way.

Then .... I had a an epiphany.

In case `_WD_Option("Sleep", .....` it can be switched to another function on every singl line of code, at any place.

So I do a fix soon.


Get rid of Local $bDemoSleep = False
> * Need to restore prior `_WD_Option("Sleep")` when exiting DemoSleep function

Ok

mlipok:
> > * Otherwise, this will affect all demo functions and not just DemoSleep
> 
> Affect ? Yes. Is it a problem ? 

Now I understand that you wanted this functionality to be decoupled as non-standard, so that the `Setup ....` functions did not imply the use of Sleep as a standard procedure.

Do I understand your intentions well?

Danp2:
> Now I understand that you wanted this functionality to be decoupled as non-standard, so that the Setup .... functions did not imply the use of Sleep as a standard procedure.
> 
> Do I understand your intentions well?

Correct, although maybe not the "non-standard" part. If the goal is to demonstrate the `_WD_Option("Sleep")` in a single function, then it should be implemented in a way that won't have any effect on the other functions.

A different option would be to have a checkbox in the GUI that would control if this functionality was enabled for all routines. But that might require each Demo function to be rewritten so that they exit whenever Abort is clicked.

Danp2:
This is more in line with what I was hoping for. 👍 

One issue I observed is that the script no longer exits if you click the X on the GUI when DemoSleep is running. I'm guessing that the $GUI_EVENT_CLOSE is being "eaten" in _USER_WD_Sleep. Thoughts?

mlipok:
> One issue I observed is that the script no longer exits if you click the X on the GUI when DemoSleep is running.

Normally when something is going wrong then you want to easily stop automation process.
Then the X button is the first click user is trying to do even when you have BIG RED BUTTON "ABORT".

So the last commit showing different `MsgBox()` title "Demo aborted!" in case user click `Abort` or `X` button.

Danp2:
What happens when `DemoSleep` isn't the last function called? Seems like `$iError` will only hold the error code from the most recent function call. I think you need to check `@error = $_WD_ERROR_UserAbort` and then set a separate flag for later use.

mlipok:
On Abort processing all other/further Demo's function should be canceled.
This is natural way how any Abort feature should behave.

Danp2:
You should be able to eliminate $iError in DemoSleep (just use @error).

mlipok:
There was wrong `Return` in `DemoSleep` 
WIP

mlipok:
Now when you way little longer then `DemoSleep()` end with success.

PR #109, opened 2021-12-15
wd_demo.au3 + GUISetBkColor

mlipok:
On my Windows 11 I do no see differences between checked and unchecked Buttons, by adding GUISetBkColor() it becomes more contrasting / distinguishable.


wd_demo.au3
```diff
@@ -55,7 +56,7 @@ Func _WD_Demo()
 	Local $aCheckboxes[$iCount]
 
 	Local $hGUI = GUICreate("Webdriver Demo", 200, 150 + (20 * $iCount), 100, 200, BitXOR($GUI_SS_DEFAULT_GUI, $WS_MINIMIZEBOX))
-
+	GUISetBkColor($COLOR_LIGHTSTEELBLUE)
```
Danp2, line 59: This doesn't compile for me. Are you running beta? I don't see this constant in ColorConstants.au3

Danp2 (changes requested):

wd_demo.au3
Danp2: Confirmed that $COLOR_LIGHTSTEELBLUE is available in latest beta, but we want this to work without requiring beta. Can you retest using a different color, such as $CLR_SILVER or $CLR_MEDGRAY?

mlipok:
btw. I find an issue when I was trying different colors then I found that some for example: $CLR_MEDGRAY has been removed form Beta... (at least from dev version of AutoIt Include).

mlipok:
current:
image

proposal:
image

PR #111, opened 2021-12-22
Add check command in _WD_ElementActionEx

TheDcoder:
Hi, I have added support for new `check` and `uncheck` commands in the `_WD_ElementActionEx` function. Please review the code and I'll amend it as appropriate for merge :)

### DIFF
--- a/wd_helper.au3
+++ b/wd_helper.au3
@@ -1731,6 +1731,7 @@ EndFunc   ;==>_WD_SetElementValue
 ;                  $sElement   - Element ID from _WD_FindElement
 ;                  $sCommand   - one of the following actions:
 ;                  |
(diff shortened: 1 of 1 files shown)


TheDcoder:
This allows setting a checkbox to a known state, and also since this uses JS it doesn't have the limitations of `_WD_Click`.

mlipok:
What kind of limitations you mean ?

TheDcoder:
By limitations I don't mean the function's code, but the limitations of the WebDriver itself in how it executes this function. I am not sure about other browsers, but Chrome doesn't like clicking on elements if they are obscured or overlapped by another element or if they aren't rendered to be visible in the page.

Using JavaScript entirely avoids this class of errors as it doesn't care about that at all :grin:

PR #115, opened 2022-01-01
_WD_ExecuteScript  $bValueNode > $sJSONNode

mlipok:


mlipok:
by mistake I do the last commit in:
https://github.com/Danp2/WebDriver/pull/114
https://github.com/Danp2/WebDriver/pull/114/commits/a7fe5b1837f8b5b4b6f275947ac84b3dd14faa9a

Danp2 (changes requested):
I don't like the way this is implemented. Instead of $vValueNode, I would suggest $iReturnType. Then we need an enum with JSON (default), value node, etc.

Danp2:
> $vValueMode

I didn't initially catch that this changed from "Node" to "Mode", but I still stand by my earlier comments. 😉

mlipok:
> I don't like the way this is implemented. Instead of $vValueNode, I would suggest $iReturnType. Then we need an enum with JSON (default), value node, etc.

ok. Will work on it

mlipok:
maybe just  this way:
`
_WD_ExecuteScript($sSession, $sScript2, $_WD_EmptyDict, Default, "[value]")
`
or
`
_WD_ExecuteScript($sSession, $sScript2, $_WD_EmptyDict, Default, "[value][" & $_WD_ELEMENT_ID & "]")
`

I mean if string is given then it means that you want to get value ...

mlipok:
If you accept please do not merge as I will need to change `_WD_GetElementFromPoint`
`
		$aCoords = _WD_ExecuteScript($sSession, $sScript2, $_WD_EmptyDict, Default, True)
`
And I want to do this here.

mlipok:
> I will need to change `_WD_GetElementFromPoint` 
Done

wd_core.au3
```diff
@@ -811,13 +811,13 @@ EndFunc   ;==>_WD_ElementAction
...
-; Syntax ........: _WD_ExecuteScript($sSession, $sScript[, $sArguments = Default[, $bAsync = Default[, $vValueMode = Default]]])
+; Syntax ........: _WD_ExecuteScript($sSession, $sScript[, $sArguments = Default[, $bAsync = Default[, $sGetValue = Default]]])
 ; Parameters ....: $sSession   - Session ID from _WD_CreateSession
 ;                  $sScript    - Javascript command(s) to run
 ;                  $sArguments - [optional] String of arguments in JSON format
 ;                  $bAsync     - [optional] Perform request asyncronously? Default is False
-;                  $vValueMode - [optional] Return value node from Webdriver response. Default is False. True returns [value]. $_WD_ELEMENT_ID returns $sElement
-; Return values .: Success - Raw response from web driver or value node.
+;                  $sGetValue  - [optional] Return value according to given string instead Raw response. Default is ""
+; Return values .: Success - Raw response from web driver or value requested by $sGetValue
```
Danp2, line 820: Rather than `value`, I would suggest `JSON node`

Danp2 (changes requested):

wd_core.au3
mlipok: $sGetValue > $sJSONNode

wd_core.au3
```diff
@@ -811,13 +811,13 @@ EndFunc   ;==>_WD_ElementAction
...
 ; Description ...: Execute Javascipt commands.
-; Syntax ........: _WD_ExecuteScript($sSession, $sScript[, $sArguments = Default[, $bAsync = Default[, $bValueNode = False]]])
+; Syntax ........: _WD_ExecuteScript($sSession, $sScript[, $sArguments = Default[, $bAsync = Default[, $sJSONNode = Default]]])
 ; Parameters ....: $sSession   - Session ID from _WD_CreateSession
 ;                  $sScript    - Javascript command(s) to run
 ;                  $sArguments - [optional] String of arguments in JSON format
 ;                  $bAsync     - [optional] Perform request asyncronously? Default is False
-;                  $bValueNode - [optional] Return value node from Webdriver response. Default is False.
-; Return values .: Success - Raw response from web driver or value node.
+;                  $sJSONNode  - [optional] Return value according to given $sJSONNode instead raw response. Default is ""
```
Danp2, line 819: This doesn't read well to me. Maybe change to something like the following instead --

> Return the designated JSON node instead of the entire JSON string

mlipok: done

Danp2 (approved):

TheDcoder:
I know that this is merged, but I would like to bring up something, does the WebDriver ever return anything other than a object with a single `value` property? I quickly glanced through the W3C draft and it doesn't mention anything other than returning the return value.

Also I think it would be better if we maintain backward compatibility with the old function and check if the argument is a boolean, and if it is and it's `True` then we treat it like before and return the value directly.

mlipok:
> I know that this is merged, but I would like to bring up something, does the WebDriver ever return anything other than a object with a single `value` property? I quickly glanced through the W3C draft and it doesn't mention anything other than returning the return value.

I do not know what you want to propose.

> Also I think it would be better if we maintain backward compatibility with the old function and check if the argument is a boolean, and if it is and it's `True` then we treat it like before and return the value directly.

IMHO current features is more flexible, and consistent.

TheDcoder:
> I do not know what you want to propose.

I am not really proposing anything, was just curious if there was an use-case for this feature, I am not aware of JSON values that might be returned by ExecuteScript. Do you have any examples of such values that you can quickly share? All of the JSON I saw follows the same format: `{value:"<insert return value here>"}`

> IMHO current features is more flexible, and consistent.

I agree, but we can also make it backward compatible for older scripts without compromising on the current new feature. Win-win for everyone.

mlipok:
> Do you have any examples of such values that you can quickly share?

Almost there... I will do next PR soon which will fit:
https://github.com/Danp2/WebDriver/issues/116

Danp2:
@TheDcoder 

>  does the WebDriver ever return anything other than a object with a single value property? I quickly glanced through the W3C draft and it doesn't mention anything other than returning the return value.

I think you may be correct. However, the idea here is to allow for the retrieval of specific nodes within the value node. For example, this is from _WD_GetMouseElement --

```
	$sResponse = _WD_ExecuteScript($sSession, $sScript, '')
	$oJSON = Json_Decode($sResponse)
	$sElement = Json_Get($oJSON, "[value][" & $_WD_ELEMENT_ID & "]")
```

With the revision, this could be written as a single line of code --

`$sElement = _WD_ExecuteScript($sSession, $sScript, '', Default, "[value][" & $_WD_ELEMENT_ID & "]")`

> Also I think it would be better if we maintain backward compatibility with the old function and check if the argument is a boolean, and if it is and it's True then we treat it like before and return the value directly.

Personally, I'm ok with not being backward compatible since this was only recently added.

mlipok:
> I'm ok with not being backward compatible since this was only recently added.

Exactly this was released officially 5 days ago.
So I do not think that this is a ScriptBreakingChange as yet.

mlipok:
> With the revision, this could be written as a single line of code --
> 
> `$sElement = _WD_ExecuteScript($sSession, $sScript, '', Default, "[value][" & $_WD_ELEMENT_ID & "]")`

Exactly.
And this will be done in relation to:
https://github.com/Danp2/WebDriver/issues/116

TheDcoder:
My bad, looks like I didn't properly understand the code first time. Now it makes more sense to me.

Although I must admit that I liked the `$bValue` parameter because it was so simple, in fact I was able to add it in myself but then saw that @mlipok had already done it. I did not realize that it was a very recent change.

Nevertheless I would love to see the parameter also accept `True` to simply return the parsed value, I can see myself using this all the time to get strings without additional processing or writing `[value]` everytime.

Danp2:
> > Do you have any examples of such values that you can quickly share?
> 
> Almost there... I will do next PR soon which will fit: #116
> 
> as currently I'm waiting for: #117 aprroval.

Approved and merged. Hopefully, you will one day be able to handle more than on PR at a time. 😝

Danp2:
> Nevertheless I would love to see the parameter also accept True to simply return the parsed value, I can see myself using this all the time to get strings without additional processing or writing [value] everytime.

Makes sense. I think @mlipok will address this for you.

Danp2:
> I don't like the way this is implemented. Instead of $vValueNode, I would suggest $iReturnType. Then we need an enum with JSON (default), value node, etc.

I still think this is how I would do it.

mlipok:
> I still think this is how I would do it.

The currently implemented way ?

Danp2:
@mlipok No, the way I quoted above with predefined enums.

mlipok:
> @mlipok No, the way I quoted above with predefined enums.

But as far as I understand you , and after checking UDF I think that this wil require 2 additional parameters, 
one to passs the enum and second to pass string for `Json_Get($oJSON, $sJSONNode)`

mlipok:
> Nevertheless I would love to see the parameter also accept `True` to simply return the parsed value, I can see myself using this all the time to get strings without additional processing or writing `[value]` everytime.

it is for you:
https://github.com/Danp2/WebDriver/pull/122/commits/87e9adaa9ba6f9e63ec5c4d79a69747027c0a17a

Danp2:
> But as far as I understand you , and after checking UDF I think that this wil require 2 additional parameters,
one to passs the enum and second to pass string for Json_Get($oJSON, $sJSONNode)

No, I meant to replace the current $sJSONNode with  $iReturnType, which would be the parameter where the enum was specified. There would be an enum for each of the allowed options (Raw, Value, Element, etc)

mlipok:
so How would you like to put parameters like this: 

`"[value][" & $_WD_ELEMENT_ID & "]"`

Which is used:
```
Func _WD_GetMouseElement($sSession)
	Local Const $sFuncName = "_WD_GetMouseElement"
	Local $sScript = "return Array.from(document.querySelectorAll(':hover')).pop()"
	Local $sElement = _WD_ExecuteScript($sSession, $sScript, '', Default, "[value][" & $_WD_ELEMENT_ID & "]")
	Local $iErr = @error

		__WD_ConsoleWrite($sFuncName & ': ' & $sElement & @CRLF)
	EndIf

	Return SetError(__WD_Error($sFuncName, $iErr), 0, $sElement)
EndFunc   ;==>_WD_GetMouseElement

```

Danp2:
> 	Local $sElement = _WD_ExecuteScript($sSession, $sScript, '', Default, "[value][" & $_WD_ELEMENT_ID & "]")

Instead of this, you would use one of the predefined enums to indicate the result that you want returned, ie --

```
Global Enum _
		$_WD_RETURN_Raw = 0, _
		$_WD_RETURN_Value = 1, _
		$_WD_RETURN_Element = 2

	Local $sElement = _WD_ExecuteScript($sSession, $sScript, '', Default, $_WD_RETURN_Element)
```

_WD_ExecuteScript will internally know how to handle the enum and return the desired result. I feel this is while not as powerful as the current implementation, it is more user friendly and will suffice for 99% of all usage.

And we could still check for True and handle it like $_WD_RETURN_Value

mlipok:
So I would propose:
```
Global Const $_WD_RETURN_Raw = False
Global Const $_WD_RETURN_Value = True
Global Const $_WD_RETURN_Element = "[value][" & $_WD_ELEMENT_ID & "]")
```
and all kind of coding ways should be possible:

Raw
```
Local $sElement = _WD_ExecuteScript($sSession, $sScript, '', Default, $_WD_RETURN_Raw)
Local $sElement = _WD_ExecuteScript($sSession, $sScript, '', Default, False)
Local $sElement = _WD_ExecuteScript($sSession, $sScript, '', Default, Default)
Local $sElement = _WD_ExecuteScript($sSession, $sScript, '', Default, '')

```

Value
```
Local $sElement = _WD_ExecuteScript($sSession, $sScript, '', Default, $_WD_RETURN_Value)
Local $sElement = _WD_ExecuteScript($sSession, $sScript, '', Default, True)
Local $sElement = _WD_ExecuteScript($sSession, $sScript, '', Default, "[value]')

```
element 
```
Local $sElement = _WD_ExecuteScript($sSession, $sScript, '', Default, $_WD_RETURN_Element)
Local $sElement = _WD_ExecuteScript($sSession, $sScript, '', Default, "[value][" & $_WD_ELEMENT_ID & "]")
```

by hand:
`Local $sElement = _WD_ExecuteScript($sSession, $sScript, '', Default, "[value][error]")

`

Danp2:
Personally, I would prefer to see 

```
Global Const $_WD_ESRETURN_Raw = ""
Global Const $_WD_ESRETURN_Value = "[value]"
Global Const $_WD_ESRETURN_Element = "[value][" & $_WD_ELEMENT_ID & "]")
```

However, it may make more sense to define these as $_WD_JSON_* so that they can be reused in other portions of the UDF.
Either way, this will still require $sJSONNode to be renamed so that it appropriately reflects the type and usage.

mlipok:
Of course internally `True` should be still valid.
@TheDcoder  Your opionin ?

TheDcoder:
I like the approach of using constants instead of arbitrary `enums`, they provide more flexibility without compromise. As for the actual values of the constants, I am leaning towards this:
```
Global Const $_WD_RETURN_Raw = False
Global Const $_WD_RETURN_Value = True
```
Simply because a boolean value is simpler than a string, we have to check both types anyway and checking for boolean by default would be easier. It self-documents the fact that `True` and `False` are valid values, otherwise we may have to specify that they're valid in the function header docs.

Danp2:
While boolean values are being supported today, that may not be true forever. Therefore, I'd suggest sticking with the string constants.

TheDcoder:
> that may not be true forever

Why not? As I have expressed previously, I would prefer using them instead of constants.

Danp2:
IDK. It just doesn't feel "natural" to me to design a function to accept multiple variable types when it isn't necessary IMO.

TheDcoder:
I thought it was one of the highlights of a dynamic-typed language, you can all types with any value :grinning:

Anyway, I know that it's a matter of opinion, I have placed mine before you and as the project leader you have the final say :)

mlipok:
> it may make more sense to define these as $_WD_JSON_* so that they can be reused in other portions of the UDF

Ok


> Either way, this will still require $sJSONNode to be renamed so that it appropriately reflects the type and usage.

Yours proporsal for the name is....

Danp2:
I was just thinking about that. My initial idea is to change the name to $vSubNode. Thoughts or other suggestions?

mlipok:
ok
I will introduce an amendment

PR #122, opened 2022-01-02
wd_helper.au3 _WD_ExecuteScript() $vSubNode usage

mlipok:
In relation to:
https://github.com/Danp2/WebDriver/pull/115
and
https://github.com/Danp2/WebDriver/issues/116

Now it is less about 48 lines

### DIFF
--- a/wd_core.au3
+++ b/wd_core.au3
@@ -862,12 +862,12 @@ Func _WD_ExecuteScript($sSession, $sScript, $sArguments = Default, $bAsync = Def
 		EndIf
 
 		If $iErr = $_WD_ERROR_Success Then
-			If StringLen($vSubNode) then
+			If StringLen($vSubNode) Then
 				Local $oJSON = Json_Decode($sResponse)
 				$sResponse = Json_Get($oJSON, $vSubNode)
 				If @error Then
 					$iErr = $_WD_ERROR_RetValue
-				Endif
+				EndIf
 			EndIf
 		EndIf
 	Else
--- a/wd_core.au3
+++ b/wd_core.au3
@@ -76,7 +76,6 @@ Global Const $_WD_LOCATOR_ByLinkText = "link text"
 Global Const $_WD_LOCATOR_ByPartialLinkText = "partial link text"
 Global Const $_WD_LOCATOR_ByTagName = "tag name"
 
-Global Const $_WD_JSON_RawResponse = ""
 Global Const $_WD_JSON_Value = "[value]"
 Global Const $_WD_JSON_Element = "[value][" & $_WD_ELEMENT_ID & "]"
 Global Const $_WD_JSON_Shadow = "[value][" & $_WD_SHADOW_ID & "]"
@@ -822,7 +821,7 @@ EndFunc   ;==>_WD_ElementAction
 ;                  $sScript    - Javascript command(s) to run
 ;                  $sArguments - [optional] String of arguments in JSON format
 ;                  $bAsync     - [optional] Perform request asyncronously? Default is False
-;                  $vSubNode  - [optional] Return the designated JSON node instead of the entire JSON string. Default is $_WD_JSON_RawResponse an raw response from web driver
+;                  $vSubNode  - [optional] Return the designated JSON node instead of the entire JSON string. Default is "" which lead to return entire response from web driver in JSON format
 ; Return values .: Success - Response from web driver in JSON format or value requested by given $vSubNode
 ;                  Failure - Response from web driver in JSON format and sets @error to value returned from __WD_Post()
 ;                            If script is executed successfully but $vSubNode isn't found, then "" (empty string) and sets @error to $_WD_ERROR_RetValue
@@ -839,9 +838,8 @@ Func _WD_ExecuteScript($sSession, $sScript, $sArguments = Default, $bAsync = Def
 
 	If $sArguments = Default Then $sArguments = ""
 	If $bAsync = Default Then $bAsync = False
-	If $vSubNode = Default Then $vSubNode = $_WD_JSON_RawResponse
-
-	If IsBool($vSubNode) Then $vSubNode = ($vSubNode) ? $_WD_JSON_Value : $_WD_JSON_RawResponse
+	If $vSubNode = Default Then $vSubNode = "" ; Response from web driver in JSON format
+	If IsBool($vSubNode) Then $vSubNode = ($vSubNode) ? $_WD_JSON_Value : "" ; Return JSON value node or entire response from web driver in JSON format
 
 	If IsString($vSubNode) Then
 		$sScript = __WD_EscapeString($sScript)
--- a/wd_core.au3
+++ b/wd_core.au3
@@ -76,6 +76,9 @@ Global Const $_WD_LOCATOR_ByLinkText = "link text"
 Global Const $_WD_LOCATOR_ByPartialLinkText = "partial link text"
 Global Const $_WD_LOCATOR_ByTagName = "tag name"
 
+Global Const $_WD_JSON_Value = "[value]"
+Global Const $_WD_JSON_Element = "[value][" & $_WD_ELEMENT_ID & "]"
+
 Global Enum _
 		$_WD_DEBUG_None = 0, _ ; No logging to console
 		$_WD_DEBUG_Error, _    ; Error logging to console
@@ -258,7 +261,7 @@ Func _WD_Status()
 
 	If $iErr = $_WD_ERROR_Success Then
 		Local $oJSON = Json_Decode($sResponse)
-		$oResult = Json_Get($oJSON, "[value]")
+		$oResult = Json_Get($oJSON, $_WD_JSON_Value)
 	EndIf
 
 	If $_WD_DEBUG = $_WD_DEBUG_Info Then
@@ -299,7 +302,7 @@ Func _WD_GetSession($sSession)
 
 	If $iErr = $_WD_ERROR_Success Then
 		Local $oJSON = Json_Decode($sResponse)
-		$sResult = Json_Get($oJSON, "[value]")
+		$sResult = Json_Get($oJSON, $_WD_JSON_Value)
 	EndIf
 
 	If $_WD_DEBUG = $_WD_DEBUG_Info Then
@@ -439,7 +442,7 @@ Func _WD_Action($sSession, $sCommand, $sOption = Default)
 
 			If $iErr = $_WD_ERROR_Success Then
 				$oJSON = Json_Decode($sResponse)
-				$sResult = Json_Get($oJSON, "[value]")
+				$sResult = Json_Get($oJSON, $_WD_JSON_Value)
 			EndIf
 
 		Case 'actions'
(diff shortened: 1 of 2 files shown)


wd_core.au3
```diff
@@ -818,7 +818,7 @@ EndFunc   ;==>_WD_ElementAction
 ;                  $bAsync     - [optional] Perform request asyncronously? Default is False
 ;                  $sJSONNode  - [optional] Return the designated JSON node instead of the entire JSON string. Default is ""
 ; Return values .: Success - Raw response from web driver or value requested by given $sJSONNode
-;                  Failure - "" (empty string) and sets @error to one of the following values:
+;                  Failure - "" (empty string) and set @error to $_WD_ERROR_RetValue, or raw response from web driver and sets @error to one of the following values:
```
Danp2, line 821: This is getting too complicated in my opinion. Is there a reason that we can't simply return an empty string in all cases? What value comes from returning the non-empty JSON response when there is an error (I haven't checked the code on this myself)?

Also, you changed `sets` to `set`. What that intentional?

wd_core.au3
```diff
@@ -1354,8 +1354,8 @@ EndFunc   ;==>__WD_Get
 ; Syntax ........: __WD_Post($sURL, $sData)
 ; Parameters ....: $sURL  - Location to access via WinHTTP
 ;                  $sData - String representing data to be sent
-; Return values..: Success - Response from web driver.
-;                  Failure - Response from web driver, sets @error to one of the following values:
+; Return values..: Success - Raw response from web driver.
```
Danp2, line 1357: I don't think adding the word `raw` adds any value here. Maybe it would be better to say "Response from web driver in JSON format" or something similar.

Danp2 (changes requested):

wd_core.au3
mlipok: > Also, you changed `sets` to `set`. What that intentional?

my fault, will change.

As to question about:
`Failure - "" (empty string) and set @error to $_WD_ERROR_RetValue, or raw response from web driver and sets @error to one of the following values:`

this part
`raw response from web driver and sets @error to one of the following values:`

comes from
`$sResponse = __WD_Post($_WD_BASE_URL & ":" & $_WD_PORT & "/session/" & $sSession & "/execute/" & $sCmd, $sData)`

and
`"" (empty string) and set @error to $_WD_ERROR_RetValue`
comes from
`$sResponse = Json_Get($oJSON, $sJSONNode)`

wd_core.au3
mlipok: This is not always JSON format, for example take a look:

```
			If $iErr Then
				$iResult = $_WD_ERROR_SendRecv
				$sResponseText = $WD_WinHTTPTimeoutMsg
			Else

```

wd_core.au3
mlipok: `Also, you changed sets to set. What that intentional?`

I checked and found that 'sets' was provided by this 
https://github.com/Danp2/WebDriver/commit/5267f95a84e4ae07106a8c5104ebc686fadd03d9

wd_core.au3
mlipok: But I get your point. Maybe you some further proposals ?

mlipok:
I wait for aproval on our discissuion here:
https://github.com/Danp2/WebDriver/pull/115#issuecomment-1005083587

Then I will make the appropriate new commitments.

Danp2:
@mlipok @TheDcoder Please review my latest commits and provide feedback.

mlipok:
I need at least 24h

mlipok:
I was able to take a look faster.
For me now all looks ok.

TheDcoder:
The changes look good to me, I have no further comments :+1:

wd_core.au3
mlipok: `Should I change all the .... sets @error to .... set @error ?`
@Danp2  what is your's decision ?

wd_core.au3
mlipok: `"Response from web driver in JSON format"`
Do I should use this `template` in entire UDF ?

wd_core.au3
Danp2: I would say no. IMO `sets` is the correct wording in the instance because we are implicitly stating that the function does something. Therefore, "the function sets @error" is correct tense where as "the function set @error" isn't.

wd_core.au3
Danp2: I think that should be fine.

Danp2 (approved):

wd_core.au3
mlipok: Sorry for the gross lack of my language skills:

Fine is current version or whether we should start using consistent nomenclature

Because we currently have 3 different forms

`Raw response from web driver`
`Response from web driver`
`Response from web driver in JSON format`

Danp2: I think ` Response from web driver in JSON format` is the best of that group

mlipok: ok

mlipok:
Now all is clear and fine to me.
As we do not change code only descriptions if you agree with mine last commit, maybe we can finally aprove and merge this PR ?

wd_core.au3
```diff
@@ -76,8 +76,11 @@ Global Const $_WD_LOCATOR_ByLinkText = "link text"
 Global Const $_WD_LOCATOR_ByPartialLinkText = "partial link text"
 Global Const $_WD_LOCATOR_ByTagName = "tag name"
 
+Global Const $_WD_JSON_RawResponse = ""
```
Danp2, line 79: I don't see why this is needed.

wd_core.au3
```diff
@@ -819,7 +822,7 @@ EndFunc   ;==>_WD_ElementAction
 ;                  $sScript    - Javascript command(s) to run
 ;                  $sArguments - [optional] String of arguments in JSON format
 ;                  $bAsync     - [optional] Perform request asyncronously? Default is False
-;                  $vSubNode  - [optional] Return the designated JSON node instead of the entire JSON string. Default is "" which returns the entire response from web driver
+;                  $vSubNode  - [optional] Return the designated JSON node instead of the entire JSON string. Default is $_WD_JSON_RawResponse an raw response from web driver
```
Danp2, line 825: I prefer the prior version (I previously mentioned that adding "raw" wasn't adding any value IMO)

Danp2 (changes requested):

wd_core.au3
mlipok: This is the same kind of need as in case `$_WD_ERROR_Success = 0` and `$_WD_DEBUG_None = 0`
Just for not using Magic Numbers

wd_core.au3
mlipok: I would prospose small change
`Default is $_WD_JSON_RawResponse which returns the entire response in JSON format from web driver`

wd_core.au3
Danp2: IMO, an empty string isn't the same thing as a magic number. This doesn't add any value; it just adds obscurity. So I can't see approving this addition.

mlipok: So why this following variables are delcared?
```
Global $_WD_DRIVER = "" ; Path to web driver executable
Global $_WD_DRIVER_PARAMS = "" ; Parameters to pass to web driver executable
Global $_WD_SESSION_DETAILS = "" ; Response from _WD_CreateSession

```

mlipok: > This doesn't add any value; it just adds obscurity.

Why `obscurity`
I propose such code:
```
	If $vSubNode = Default Then $vSubNode = $_WD_JSON_RawResponse

	If IsBool($vSubNode) Then $vSubNode = ($vSubNode) ? $_WD_JSON_Value : $_WD_JSON_RawResponse

	If IsString($vSubNode) Then

```

without $_WD_JSON_RawResponse it should look like following snippet:
```
	If $vSubNode = Default Then $vSubNode = ""

	If IsBool($vSubNode) Then $vSubNode = ($vSubNode) ? $_WD_JSON_Value : ""

	If IsString($vSubNode) Then
```
Now reading the code my understanding of the code is smaller.

So maybe like this:
```
	If $vSubNode = Default Then $vSubNode = "" ; Return raw JSON response from web driver
	If IsBool($vSubNode) Then $vSubNode = ($vSubNode) ? $_WD_JSON_Value : "" ; Return JSON value or raw JSON response from web driver

	If IsString($vSubNode) Then
```

Then the code will not contain this constant, but it will be easier to know how it works

Danp2: Yes, that's better. But remember that the header already defines that "" = returns the entire response in JSON format from web driver. Also, I would prefer this --
```
	If IsBool($vSubNode) Then $vSubNode = ($vSubNode) ? $_WD_JSON_Value : "" ; Return JSON value node if $vSubNode = True
```

mlipok: so take a look at: `$_WD_JSON_RawResponse removal` commit

wd_core.au3
```diff
@@ -839,9 +838,8 @@ Func _WD_ExecuteScript($sSession, $sScript, $sArguments = Default, $bAsync = Def
 
 	If $sArguments = Default Then $sArguments = ""
 	If $bAsync = Default Then $bAsync = False
-	If $vSubNode = Default Then $vSubNode = $_WD_JSON_RawResponse
-
-	If IsBool($vSubNode) Then $vSubNode = ($vSubNode) ? $_WD_JSON_Value : $_WD_JSON_RawResponse
+	If $vSubNode = Default Then $vSubNode = "" ; Response from web driver in JSON format
+	If IsBool($vSubNode) Then $vSubNode = ($vSubNode) ? $_WD_JSON_Value : "" ; Return JSON value node or entire response from web driver in JSON format
```
Danp2, line 842: I think this could be further defined, ie --

```
; When $vSubNode is True, configure so that the JSON value node is returned.
; When $vSubNode is False, configure so that the entire response is returned.
```

Danp2 (changes requested):

wd_core.au3
```diff
@@ -822,7 +821,7 @@ EndFunc   ;==>_WD_ElementAction
 ;                  $sScript    - Javascript command(s) to run
 ;                  $sArguments - [optional] String of arguments in JSON format
 ;                  $bAsync     - [optional] Perform request asyncronously? Default is False
-;                  $vSubNode  - [optional] Return the designated JSON node instead of the entire JSON string. Default is $_WD_JSON_RawResponse an raw response from web driver
+;                  $vSubNode  - [optional] Return the designated JSON node instead of the entire JSON string. Default is "" which lead to return entire response from web driver in JSON format
```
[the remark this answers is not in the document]
mlipok: done

wd_core.au3
mlipok: Simler:
`; True = the JSON value node is returned , False entire JSON response is returned`
?

wd_core.au3
```diff
@@ -839,9 +838,8 @@ Func _WD_ExecuteScript($sSession, $sScript, $sArguments = Default, $bAsync = Def
 
 	If $sArguments = Default Then $sArguments = ""
 	If $bAsync = Default Then $bAsync = False
-	If $vSubNode = Default Then $vSubNode = $_WD_JSON_RawResponse
-
-	If IsBool($vSubNode) Then $vSubNode = ($vSubNode) ? $_WD_JSON_Value : $_WD_JSON_RawResponse
+	If $vSubNode = Default Then $vSubNode = "" ; Response from web driver in JSON format
```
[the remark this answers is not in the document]
mlipok: Ok

Danp2 (approved):

PR #123, opened 2022-01-02
wd_helper.au3 - _WD_GetFrameCount - Remarks

mlipok:


wd_helper.au3
```diff
@@ -490,14 +490,14 @@ EndFunc   ;==>_WD_LastHTTPResult
 
 ; #FUNCTION# ====================================================================================================================
 ; Name ..........: _WD_GetFrameCount
-; Description ...: This will return how many frames/iframes are in your current window/frame.
+; Description ...: This will return how many frames/iframes are in your current context (window/frame/document).
 ; Syntax ........: _WD_GetFrameCount($sSession)
 ; Parameters ....: $sSession - Session ID from _WD_CreateSession
-; Return values .: Success - Numeric count of frames, 0 or positive number.
+; Return values .: Success - Number of frames, 0 means that no frames were found in the current context
```
Danp2, line 496: Not sure that the explanation for the meaning of "0" is necessary. What else could it possibly mean?

wd_helper.au3
```diff
@@ -490,14 +490,14 @@ EndFunc   ;==>_WD_LastHTTPResult
 
 ; #FUNCTION# ====================================================================================================================
 ; Name ..........: _WD_GetFrameCount
-; Description ...: This will return how many frames/iframes are in your current window/frame.
+; Description ...: This will return how many frames/iframes are in your current context (window/frame/document).
```
Danp2, line 493: What's the difference between window and document in this context? Are they truly different?

wd_helper.au3
```diff
@@ -490,14 +490,14 @@ EndFunc   ;==>_WD_LastHTTPResult
...
 ; Syntax ........: _WD_GetFrameCount($sSession)
 ; Parameters ....: $sSession - Session ID from _WD_CreateSession
-; Return values .: Success - Numeric count of frames, 0 or positive number.
+; Return values .: Success - Number of frames, 0 means that no frames were found in the current context
 ;                  Failure - "" (empty string) and sets @error to $_WD_ERROR_Exception
 ; Author ........: Decibel, Danp2
 ; Modified ......: 2018-04-27
-; Remarks .......: The function will not traverse to nested frames
+; Remarks .......: The function will not traverse to nested frames - it will count frames only from the current context (window/frame/document)
```
Danp2, line 500: This is partial redundant with the changed description above. It probably could just state `Nested frames are not included in the frame count`

Danp2 (changes requested):

wd_helper.au3
mlipok: done

wd_helper.au3
mlipok: done

wd_helper.au3
mlipok: in JavaScript you can use something like this: `window.frames[i].document.body.style.background = "red";` and that mean when you are inside of any frame then you are on Document. And thus my recent commit - please check them.

Danp2 (approved):

PR #127, opened 2022-01-09
_WD_UpdateDriver - revised

mlipok:
Today I got problems with _WD_UpdateDriver() and I do not know where the issue was.
So I do some checking and thus my update.
There was redundant checking for @error in 
```
		$sDriverVersion = _WD_GetWebDriverVersion($sInstallDir, $sDriverEXE)
		If @error Then $iErr = @error
```
It was redundant because of:
```
			; When $bForce parameter equals Null, then return True if newer driver is available

```

and there was also lack of check for the case of `$sDriverVersion = 'None'` and thus when I have'nt already downloaded any `chromedriver.exe` or `geckodriver.exe` or `msedgedriver.exe` then this function never started downloading a newer version.
So I added them here:
`ElseIf $sDriverLatest > $sDriverVersion Or $bForce Or $sDriverVersion = 'None' Then
`

I also added some additional __WD_ConsoleWrite()


wd_helper.au3
```diff
@@ -1389,7 +1389,7 @@ Func _WD_UpdateDriver($sBrowser, $sInstallDir = Default, $bFlag64 = Default, $bF
 				If $sDriverLatest > $sDriverVersion Then
 					$bResult = True
 				EndIf
-			ElseIf $sDriverLatest > $sDriverVersion Or $bForce Then
+			ElseIf $sDriverLatest > $sDriverVersion Or $bForce Or $sDriverVersion = 'None' Then
```
Danp2, line 1392: I'd have to go back and check past discussions, but I seem to recall that I previously told others that they needed to set $bForce = True in the case where the webdriver didn't previously exist.

Danp2 (changes requested):

wd_helper.au3
Danp2: Found it. See discussion [here](https://www.autoitscript.com/forum/topic/201106-webdriver-udf-help-support-ii/?do=findComment&comment=1467771)

mlipok:
Sorry for all my today comments to this PR.
I just notice that they was in PENDING state for last few days.
I just submit them now.

wd_helper.au3
```diff
@@ -1326,7 +1326,6 @@ Func _WD_UpdateDriver($sBrowser, $sInstallDir = Default, $bFlag64 = Default, $bF
 		If @error Then $iErr = @error
 
 		$sDriverVersion = _WD_GetWebDriverVersion($sInstallDir, $sDriverEXE)
-		If @error Then $iErr = @error
```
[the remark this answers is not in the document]
Danp2: Good catch. It looks like I introduced the issue in 21d7269e620649a3ed87ee73b3edbbeec86c4d7f.

wd_helper.au3
Danp2: > so why `$bForce .... Default is False` ?
> Is this for security reason that Coders/Scripter must intentionaly download file ?

No. It's simply the way the function was designed to work.

> And the description is not accurate I mean misleading at least I was not aware about how to use them.

Can you be more specific? What would you change? We can certainly add more to the Remarks that clarify the functionality.

Danp2: > I do not want to update driver each run .
So how do you upgrade, but only if a newer version is available or no driver is available?

I see your point. In the current situation, you would need to call _WD_GetWebDriverVersion first, and then adjust your call to _WD_UpdateDriver as needed if the driver isn't present.

wd_helper.au3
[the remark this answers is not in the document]
mlipok: So what you propose here to solve this issue?
Is my fix good ?

wd_helper.au3
mlipok: > > And the description is not accurate I mean misleading at least I was not aware about how to use them.
> 
> Can you be more specific? What would you change? We can certainly add more to the Remarks that clarify the functionality.

Current description is:
`$bForce      - [optional] Force update? Default is False`

but it act like:
`$bForce      - [optional] Force download? Default is False`

because if you set True then it always download if you current local version is correct, which mean that you do not update but re download, which mean you forced to download regardles of any case. For me update mean only if there is lack of current version.

wd_helper.au3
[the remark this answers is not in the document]
Danp2: Not sure. I'll have to take a closer look at it when I have some time.

wd_helper.au3
Danp2: > Force update? vs Force download?

To me, they pretty much mean the same thing because you have to download the update in order to apply it, correct? In this scenario, how would you perform an update without downloading the file?

Also, the function name is _WD_UpdateDriver, not _WD_DownloadDriver. Thus the existing description works for me. <shrug>

Danp2:
@mlipok Take a look at the latest commits and let me know if this looks better to you. I'm still not sure about the added check for "None". I'm wondering if 

- _WD_GetWebDriverVersion should return "0.0" instead of "None"
- Should we be converting the strings to numbers before comparing?

wd_helper.au3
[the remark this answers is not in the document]
mlipok: I check your commit ASAP

Danp2:
@mlipok Can you explain why you merge main branch into PR? Is that really needed in this scenario?

wd_helper.au3
```diff
@@ -1258,143 +1258,147 @@ EndFunc   ;==>_WD_IsLatestRelease
...
 	Local $sReturned, $sTempFile, $hFile, $oShell, $FilesInZip, $sResult, $iStartPos, $iConversion
 
 	If $sInstallDir = Default Then $sInstallDir = @ScriptDir
 	If $bFlag64 = Default Then $bFlag64 = False
 	If $bForce = Default Then $bForce = False
 
 	; If the Install directory doesn't exist and it can't be created, then set error
-	If (Not FileExists($sInstallDir)) And (Not DirCreate($sInstallDir)) Then $iErr = $_WD_ERROR_InvalidValue
+	If (Not FileExists($sInstallDir)) And (Not DirCreate($sInstallDir)) Then
+		$iErr = $_WD_ERROR_InvalidValue
```
Danp2, line 1270: I feel like we should also check for write access as well in case they do something stupid like passing in "c:\windows".

wd_helper.au3
mlipok: Force update mean only when there is no dirver available or the driver is not up to date.
Force download .... regardless of the version you have - that is always.

mlipok: Current version of this function works exactly in this way.

wd_helper.au3
```diff
@@ -1343,6 +1343,9 @@ Func _WD_UpdateDriver($sBrowser, $sInstallDir = Default, $bFlag64 = Default, $bF
 			If $iErr = $_WD_ERROR_Success Then
 				Local $bDriverExists = ($sDriverCurrent <> 'None')
 
+				; Determine if the remote and local files are the same size, just in case they are different, force a download. This is needed because 32/64 bit versions have different file sizes.
+				If FileGetSize($sInstallDir & $sDriverEXE) <> InetGetSize($sURLNewDriver, $INET_FORCERELOAD) Then $bForce = True
+
```
mlipok, line 1348: This is required due to the fact that `_WinAPI_GetBinaryType()` returns `$SCS_DOS_BINARY` and you can not determine if current version is 32/64bit because `msedgedriver.exe --version` and also `chromedriver.exe --version` returns the same information for `--version` for both 32 and 64 bit versions..

@Danp2  waiting for yours approval.
I think that this will be all in this PR

mlipok:
> @mlipok Can you explain why you merge main branch into PR? Is that really needed in this scenario?

I supposed that this was required.

So I'm wondering now: If it was not required, when will it be required?

mlipok:
btw. 
I think that this PR name `_WD_UpdateDriver - better @error handling` is not accurate.
Should it be changed? Do you have any suggestion?

wd_helper.au3
```diff
@@ -1343,6 +1343,9 @@ Func _WD_UpdateDriver($sBrowser, $sInstallDir = Default, $bFlag64 = Default, $bF
 			If $iErr = $_WD_ERROR_Success Then
 				Local $bDriverExists = ($sDriverCurrent <> 'None')
 
+				; Determine if the remote and local files are the same size, just in case they are different, force a download. This is needed because 32/64 bit versions have different file sizes.
+				If FileGetSize($sInstallDir & $sDriverEXE) <> InetGetSize($sURLNewDriver, $INET_FORCERELOAD) Then $bForce = True
```
Danp2, line 1347: Wouldn't it make sense to change this to `If $bDriverExists And FileGetSize($sInstallDir & $sDriverEXE) <> InetGetSize($sURLNewDriver, $INET_FORCERELOAD) Then $bForce = True`? No point in checking the file sizes if we know that the local file doesn't exist.

wd_helper.au3
```diff
@@ -1335,6 +1334,7 @@ Func _WD_UpdateDriver($sBrowser, $sInstallDir = Default, $bFlag64 = Default, $bF
 				$sVersionShort = StringLeft($sBrowserVersion, StringInStr($sBrowserVersion, ".", 0, -1) - 1)
 				$sDriverLatest = BinaryToString(InetRead('https://chromedriver.storage.googleapis.com/LATEST_RELEASE_' & $sVersionShort))
 				$sURLNewDriver = "https://chromedriver.storage.googleapis.com/" & $sDriverLatest & "/chromedriver_win32.zip"
+				; to manually check visit this site:  https://sites.google.com/chromium.org/driver/
```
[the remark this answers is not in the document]
mlipok: I wonder why this is marked as `Outedated` as I resolve them before...
image

wd_helper.au3
mlipok: Done

mlipok:
> @mlipok Take a look at the latest commits and let me know if this looks better to you. I'm still not sure about the added check for "None". I'm wondering if
> 
> * _WD_GetWebDriverVersion should return "0.0" instead of "None"
> * Should we be converting the strings to numbers before comparing?

As far as I'm concerned, this is the last thing to check.

wd_helper.au3
Danp2: In hindsight, this won't work because you are comparing the size of the local EXE to the remote ZIP, which will never match.

Danp2 (changes requested):

wd_helper.au3
Danp2: > This is required due to the fact that _WinAPI_GetBinaryType() returns $SCS_DOS_BINARY

Are you sure about this? I got $SCS_32BIT_BINARY and $SCS_64BIT_BINARY when I tested it.

Danp2: Maybe a difference between Win7 and Win10?

mlipok: using this code:

```AutoIt
#include <WinAPIFiles.au3>
ConsoleWrite("- " & @ScriptLineNumber & " " & _WinAPI_GetBinaryType("z:\Test\WebDriver\chromedriver.exe") & @CRLF)
ConsoleWrite("- " & @ScriptLineNumber & " " & _WinAPI_GetBinaryType("z:\Test\WebDriver\geckodriver.exe") & @CRLF)
ConsoleWrite("- " & @ScriptLineNumber & " " & _WinAPI_GetBinaryType("z:\Test\WebDriver\msedgedriver.exe") & @CRLF)

ConsoleWrite("- " & @ScriptLineNumber & " " & _WinAPI_GetBinaryType("z:\Test\WebDriver\1\chromedriver.exe") & @CRLF)
ConsoleWrite("- " & @ScriptLineNumber & " " & _WinAPI_GetBinaryType("z:\Test\WebDriver\1\geckodriver.exe") & @CRLF)
ConsoleWrite("- " & @ScriptLineNumber & " " & _WinAPI_GetBinaryType("z:\Test\WebDriver\1\msedgedriver.exe") & @CRLF)
Exit

```
I always get `1`

wd_helper.au3
mlipok: reverted

wd_helper.au3
mlipok: I use Windows 11 on my notebook.

mlipok: Take a look here:
https://www.autoitscript.com/forum/topic/205553-webdriver-udf-help-support-iii/?do=findComment&comment=1495308

mlipok: Done

mlipok:
> > @mlipok Take a look at the latest commits and let me know if this looks better to you. I'm still not sure about the added check for "None". I'm wondering if
> > 
> > * _WD_GetWebDriverVersion should return "0.0" instead of "None"
> > * Should we be converting the strings to numbers before comparing?
> 
> As far as I'm concerned, this is the last thing to check.

_VersionCompare() ?

wd_helper.au3
```diff
@@ -1284,6 +1285,15 @@ Func _WD_UpdateDriver($sBrowser, $sInstallDir = Default, $bFlag64 = Default, $bF
 					$sDriverEXE = "msedgedriver.exe"
 			EndSwitch
 
+			; Determine current local webdriver Architecture
+			If FileExists($sInstallDir & $sDriverEXE) Then
+				_WinAPI_GetBinaryType($sInstallDir & $sDriverEXE)
+				Local $bDriverIs64Bit = (@extended = $SCS_64BIT_BINARY)
+				If $sBrowser = 'firefox' Or $sBrowser = 'msedge' Then
+					If $bDriverIs64Bit <> $bFlag64 Then FileDelete($sInstallDir & $sDriverEXE)
+				EndIf
```
Danp2, line 1299: I think that we should consider either
- dropping the browser check here, or
- changing it to just check <> "chrome"

wd_helper.au3
```diff
@@ -1471,7 +1471,7 @@ EndFunc   ;==>_WD_GetBrowserVersion
 ; ===============================================================================================================================
 Func _WD_GetWebDriverVersion($sInstallDir, $sDriverEXE)
 	Local Const $sFuncName = "_WD_GetWebDriverVersion"
-	Local $sDriverVersion = "None"
+	Local $sDriverVersion = 0
```
Danp2, line 1478: This is a script breaking change. I think some additional thought should go into this before proceeding.

Danp2: Also doesn't make sense to assign a numeric value to a string variable. 😛

wd_helper.au3
mlipok: > * dropping the browser check here, or

?

wd_helper.au3
mlipok: There was inconsistence with Function Header 
`Failure - "" (empty string) and sets @error to one of the following values:`
It was not true because at the top of the function it was declared as `$sDriverVersion = "None"`

wd_helper.au3
Danp2: > ?

The initial If...Endif could be viewed as unneeded as the function will still work without the "special handling" for Chrome not having a 64 bit version of the webdriver.

wd_helper.au3
mlipok: `_WD_GetWebDriverVersion()` was introduced not so long ago, and is used mainly in `_WD_Update()` 
Even if it is used by any coder/scripter should be checked with `_VersionCompare()` if it was'nt checked this way it was checked wrongly, so this is not ScriptBreaking Change, it is a fix.

Danp2: That fine. But it doesn't address my initial concerns posted above.

mlipok: > Also doesn't make sense to assign a numeric value to a string variable. 😛

changed to `$sDriverVersion = "0"`

mlipok: > I think some additional thought should go into this before proceeding.

I will wait for your thoughts ;)

mlipok: Just checked:
even if we back to `Local $sDriverVersion = "None"` then `_WD_Update()` with using `_VersionCompare()` still works well.

wd_helper.au3
mlipok: > * changing it to just check <> "chrome"

done

wd_helper.au3
Danp2: I'm ok with changing it to "0", "0.00", or whatever. We just need to clearly document the change as potentially script breaking just in case someone is checking for "none".

wd_helper.au3
```diff
@@ -1247,155 +1249,165 @@ EndFunc   ;==>_WD_IsLatestRelease
...
 
-		$sDriverVersion = _WD_GetWebDriverVersion($sInstallDir, $sDriverEXE)
-		If @error Then $iErr = @error
+			; Determine current local webdriver Architecture
+			If FileExists($sInstallDir & $sDriverEXE) Then
+				_WinAPI_GetBinaryType($sInstallDir & $sDriverEXE)
+				Local $bDriverIs64Bit = (@extended = $SCS_64BIT_BINARY)
+				If $sBrowser <> 'chrome' Then
+					If $WDDebugSave = $_WD_DEBUG_Info Then
+						__WD_ConsoleWrite($sFuncName & ': ' & $sDriverEXE & ' = ' & (($bDriverIs64Bit) ? ("switching 64>32 Bit") : ("switching 32>64 Bit")) & @CRLF)
```
Danp2, line 1295: Not sure why this is needed (there are 8 lines being output to the console, which is too much IMO), but there's something lacking here because it gets output even when no changes are being made. Perhaps you meant for this to go inside the `If $bDriverIs64Bit <> $bFlag64 Then` that follows.

wd_helper.au3
```diff
@@ -1247,155 +1249,165 @@ EndFunc   ;==>_WD_IsLatestRelease
...
+					EndIf
 
-		; Determine latest available webdriver version
-		; for the designated browser
-		Switch $sBrowser
-			Case 'chrome'
-				$sVersionShort = StringLeft($sBrowserVersion, StringInStr($sBrowserVersion, ".", 0, -1) - 1)
-				$sDriverLatest = BinaryToString(InetRead('https://chromedriver.storage.googleapis.com/LATEST_RELEASE_' & $sVersionShort))
-				$sURLNewDriver = "https://chromedriver.storage.googleapis.com/" & $sDriverLatest & "/chromedriver_win32.zip"
+					If $bDriverIs64Bit <> $bFlag64 Then FileDelete($sInstallDir & $sDriverEXE)
```
Danp2, line 1298: Do we really want to delete the driver here? There's existing code that will delete the driver, so I think you could just set `$bForce = True` here.

Danp2 (changes requested):

wd_helper.au3
mlipok: Fixed.
And about the for 8 lines
Which one you think are handy, and which are useless ?

wd_helper.au3
mlipok: Done

mlipok:
I notice that:
`_WD_UpdateDriver('msedge', @ScriptDir , False, True)`
or
`_WD_UpdateDriver('msedge', @ScriptDir , True, True)`

extracts also `Driver_Notes` directory .
Is it needed ?

wd_helper.au3
mlipok: > I'm ok with changing it to "0", "0.00", or whatever.

Let stick with 0



> We just need to clearly document the change as potentially script breaking just in case someone is checking for "none".

in REVISIONS.md ?

Danp2: Yes. I have already started the revised list locally, so I'll add it there.

Danp2 (changes requested):
Should be "0" for consistency with other function

Danp2:
> extracts also Driver_Notes directory . Is it needed ?

That comes from the MSEdge webdriver. It isn't needed, but would require changing the extraction code to only copy the desired executable.

mlipok:
> Should be "0" for consistency with other function.
> 
> Edit: Now I see that you changed the other one back to "None". Please explain.

Currentyl there is not used "None" 
It is set as `Local $sDriverVersion = 0` 

Do you mean that this should be `Local $sDriverVersion = "0"` ?

mlipok:
> It isn't needed, but would require changing the extraction code to only copy the desired executable.

I will make a relevant fix that extracts only EXE files.

Danp2:
> Currentyl there is not used "None"

I may be mistaken, but I thought that you switched _WD_GetWebDriverVersion back to "None"

> Do you mean that this should be Local $sDriverVersion = "0" ?

Yes. It's the same thing we discussed yesterday where you are assigning a numeric value to a string variable. 

IMO, both _WD_GetWebDriverVersion and _WD_GetBrowserVersion should return the same result for failure (ie: both return "None" or both return "0"). Pick one and stick to it. 😉

mlipok:
I choose
> both return "0"

because `"None"` dose'nt work with `_VersionCompare()`

mlipok:
> I choose
> 
> > both return "0"
> 
> because `"None"` doesn't work with `_VersionCompare()`

done

wd_helper.au3
mlipok: Done

Danp2 (approved):

Danp2 (approved):

PR #129, opened 2022-01-17
Cleanup

mlipok:


wd_helper.au3
```diff
@@ -1722,7 +1722,6 @@ EndFunc   ;==>_WD_SetElementValue
 ; Parameters ....: $sSession   - Session ID from _WD_CreateSession
 ;                  $sElement   - Element ID from _WD_FindElement
 ;                  $sCommand   - one of the following actions:
-;                  |
```
Danp2, line 1725: This line was there by design. It provided some separation from the prior line, which carries over to the help file as well.

mlipok: reverted

mlipok:
Will change this later with further changes.

wd_helper.au3
```diff
@@ -788,7 +788,7 @@ EndFunc   ;==>_WD_LoadWait
 ; ===============================================================================================================================
 Func _WD_Screenshot($sSession, $sElement = Default, $nOutputType = Default)
 	Local Const $sFuncName = "_WD_Screenshot"
-	Local $sResponse, $sResult, $iErr
+	Local $sResponse, $sResult = "", $iErr
```
Danp2, line 791: AFAICS, this change isn't needed because this is already accomplished in the `Else` portion of the 2nd If..Else..EndIf.

I've seen you make similar changes before. Would you agree that both coding styles are valid?

wd_helper.au3
```diff
@@ -1150,7 +1150,7 @@ EndFunc   ;==>_WD_GetShadowRoot
 Func _WD_SelectFiles($sSession, $sStrategy, $sSelector, $sFilename)
 	Local Const $sFuncName = "_WD_SelectFiles"
 
-	Local $sResult, $sJsonElement, $sSavedEscape
+	Local $sResult = "0", $sJsonElement, $sSavedEscape
```
Danp2, line 1153: Same issue here.

Danp2 (changes requested):

wd_helper.au3
```diff
@@ -1410,8 +1410,8 @@ EndFunc   ;==>_WD_UpdateDriver
 ;                  Failure - "" (empty string) and sets @error to one of the following values:
 ;                  - $_WD_ERROR_InvalidValue
 ;                  - $_WD_ERROR_NotFound
-; Author ........: Dan Pollak, mLipok
-; Modified ......: 18/06/2021
+; Author ........: Danp2
+; Modified ......: mLipok
```
Danp2, line 1414: I had always intended to used the Modified header as a date to track when it was last modified. Is there some AutoIt "standard" that says that Modified should contain the additional authors name?

wd_helper.au3
mlipok: reverting/fixing

wd_helper.au3
mlipok: no here is not the same issue becuase when `_WD_FindElement()` return Error then `$sResult` will not be set to `"0"`

```AutoIt
	Local $sElement = _WD_FindElement($sSession, $sStrategy, $sSelector)
	Local $iErr = @error

```

We can do this by moving this part:
```AutoIt
			$sJsonElement = '{"' & $_WD_ELEMENT_ID & '":"' & $sElement & '"}'
			$sResult = _WD_ExecuteScript($sSession, "return arguments[0].files.length", $sJsonElement, Default, $_WD_JSON_Value)
			$iErr = @error
			If @error Then $sResult = "0"
		Else
			$sResult = "0"
		EndIf

```

out of the main `If....EndIf` area

wd_helper.au3
mlipok: https://www.autoitscript.com/wiki/UDF-spec
```
Modified

This is a list of modifications made. At its most basic this is just a list of users who have made changes, but ideally it includes some information about what they did, in which cases it follows the same form as other multi-value lines, with the username and description of modifications seperated by a dash.

```

wd_helper.au3
Danp2: You are correct. In this case, I would stick with your original change and remove these lines --

```
		Else
			$sResult = "0"
```

Danp2 (approved):

PR #130, opened 2022-01-17
_WD_ElementActionEx() refactoring

mlipok:
This draft PR is related to:
https://github.com/Danp2/WebDriver/issues/120


mlipok:
@Danp2  what you think so far about this PR ?

Danp2:
Thus far, my opinion hasn't changed [from here](https://github.com/Danp2/WebDriver/issues/120#issuecomment-1005102924). Looking at the existing code, I do see where there could be some efficiencies by adding something like this:

```Autoit
Func __WD_BuildButtonAction($iButton, $sAction)
	Return ',{"button": ' & $iButton & ', "type": "' & $sAction & '"}' 
EndFunc
```

mlipok:
as you add here: https://github.com/Danp2/WebDriver/pull/132
`__WD_JsonElement`

I would propose to change
`__WD_BuildButtonAction` 
to
`__WD_JsonButtonAction`

or something like this.
What do you think about it ?

mlipok:
@Danp2  , @TheDcoder and @seadoggie01  please check all this JSON string reformating provided in this PR.

mlipok:
With this example:

```AutoIt
#include <WinAPIFiles.au3>
#include "..\..\wd_helper.au3"
#include "..\..\wd_capabilities.au3"
;~ #include "Z:\!!!_SVN_AU3\UDF_Forum\Other_Members\Official_AutoIt_Forum\WebDriver\wd_helper.au3"

_Example()

Func _Example()

	Local $sDesiredCapabilities = SetupChrome()
	_WD_Startup()
	If @error <> $_WD_ERROR_Success Then Return
	Local $sSession = _WD_CreateSession($sDesiredCapabilities)
	_WD_Navigate($sSession, 'https://www.google.com')
	MsgBox(0, "", 'Before click - manually accept all stuff')
	Local $sElement = _WD_FindElement($sSession, $_WD_LOCATOR_ByXPath, '//img[@class="lnXdpd"]')
	_WD_ElementActionEx($sSession, $sElement, 'CLICKANDHOLD', Default, Default, 1, 5000)
EndFunc   ;==>_Example

Func SetupChrome()
	_WD_UpdateDriver('chrome')
	_WD_Option('Driver', 'chromedriver.exe')
	_WD_Option('Port', 9515)
	_WD_Option('DriverParams', '--verbose --log-path="' & @ScriptDir & '\chrome.log"')

;~ 	Local $sDesiredCapabilities = '{"capabilities": {"alwaysMatch": {"goog:chromeOptions": {"w3c": true, "excludeSwitches": [ "enable-automation"]}}}}'
	_WD_CapabilitiesStartup()
	_WD_CapabilitiesAdd('alwaysMatch', 'chrome')
	_WD_CapabilitiesAdd('w3c', True)
	_WD_CapabilitiesAdd('excludeSwitches', 'enable-automation')
	_WD_CapabilitiesDump(@ScriptLineNumber) ; dump current Capabilities setting to console - only for testing in this demo
	Local $sDesiredCapabilities = _WD_CapabilitiesGet()
	Return $sDesiredCapabilities
EndFunc   ;==>SetupChrome

```

I have such results:

```JSON
{
	"actions": [

		{
			"id": "hover"
			,"type": "pointer"
			,"parameters": {"pointerType": "mouse"}
			,"actions": [
				{
					"type": "pointerMove"
					,"duration": 100
					,"x": 0
					,"y": 0
 					,"origin": {
 						"ELEMENT": "783b4a02-0cf1-44cf-b241-42a520ec5c09"
 						,"element-6066-11e4-a52e-4f735466cecf": "783b4a02-0cf1-44cf-b241-42a520ec5c09"
 					}
				}
,{
	"button": 1
	,"type": "pointerDown"
}
, {
	"type": "pause"
	,"duration": 5000
}
,{
	"button": 1
	,"type": "pointerUp"
}
			]
		}

	]
}

```
and as single line
```JSON
{"actions": [{"id": "hover","type": "pointer","parameters": {"pointerType": "mouse"},"actions": [{"type": "pointerMove","duration": 100,"x": 0,"y": 0 ,"origin": { "ELEMENT": "783b4a02-0cf1-44cf-b241-42a520ec5c09" ,"element-6066-11e4-a52e-4f735466cecf": "783b4a02-0cf1-44cf-b241-42a520ec5c09" }},{"button": 1,"type": "pointerDown"}, {"type": "pause","duration": 5000},{"button": 1,"type": "pointerUp"}]}]}
```

wd_helper.au3
```diff
@@ -2122,3 +2108,20 @@ EndFunc   ;==>__WD_ElementBuildActionString
...
+; Return values .: Formatted JSON string
+; Author ........: Danp2
+; Modified ......:
+; Remarks .......:
+; Related .......: _WD_ElementActionEx
+; Link ..........:
+; Example .......: No
+; ===============================================================================================================================
+Func __WD_JsonButtonAction($iButton, $sAction)
+	Return ',{"button": ' & $iButton & ', "type": "' & $sAction & '"}'
```
Danp2: @mlipok Do you think there is any value in making the comma at the beginning optional?

mlipok: There was inconsistency in JSON formating before.
for example there was:
```AutoIt
$sPostHoverAction = ',{"button":' & $iButton & ',"type":"pointerDown"},{"type": "pause", "duration": ' & $iHoldDelay & '},{"button":' & $iButton & ',"type":"pointerUp"}'
.....
$sPostAction = ',{"type": "key", "id": "keyboard_2", "actions": [{"type": "keyUp", "value": "' & $sModifier & '"}]}'

```
I just unificate them, but we can stick with a version which will not contain `space` after `:`

wd_helper.au3
```diff
@@ -1746,7 +1746,7 @@ EndFunc   ;==>_WD_SetElementValue
 ; Modified ......: TheDcoder, mLipok
 ; Remarks .......: Moving the mouse pointer above the target element is the first thing to occur for every $sCommand before it gets executed.
 ;                  There are examples in DemoElements function in wd_demo
-; Related .......: _WD_ElementAction, _WD_Action
+; Related .......: _WD_ElementAction, _WD_Action, __WD_ElementActionExStringBuilder, __WD_JsonButtonAction
```
Danp2: These don't need to be listed here IMO

wd_helper.au3
mlipok: btw. I just tested example JSON i posted above, by putting them to Notepad++ and Tidy them...
All instances of `:` was alwas suplemented with `space`.

wd_helper.au3
```diff
@@ -1783,33 +1783,24 @@ Func _WD_ElementActionEx($sSession, $sElement, $sCommand, $iXOffset = Default, $
...
 			$sPostHoverAction = _
-					',{' & _
-					'	"button": 2' & _
-					'	,"type": "pointerDown"}' & _
-					',{' & _
-					'	"button": 2' & _
-					'	,"type": "pointerUp"' & _
-					'}'
+					__WD_JsonButtonAction("2", "pointerDown") & _
+					__WD_JsonButtonAction("2", "pointerUp")
```
Danp2, line 1798: First parameter is defined as an integer, not a string. 😉

wd_helper.au3
mlipok: But this is trully not necesary and we can unificate them by removing them.
It will have positive impact, less data sent to WebDriver.

Danp2 (changes requested):

wd_helper.au3
Danp2: 1) Spaces outside of quotes are optional in JSON, so it's purely an issue of aesthetics
2) My question was regarding the comma at the beginning of the string, not spaces.

wd_helper.au3
```diff
@@ -2107,15 +2107,16 @@ EndFunc   ;==>__WD_ElementBuildActionString
...
 EndFunc   ;==>__WD_ErrHnd
+
 ; #INTERNAL_USE_ONLY# ===========================================================================================================
 ; Name ..........: __WD_JsonButtonAction
 ; Description ...: Create JSON string for button actions
 ; Syntax ........: __WD_JsonButtonAction($iButton, $sAction)
-; Parameters ....: $iButton             - an integer value.
-;                  $sAction             - a string value.
+; Parameters ....: $iButton             - an integer value declared in _WD_ElementActionEx
+;                  $sAction             - a string value declared in _WD_ElementActionEx
```
Danp2, line 2128: I don't believe this should be added. What if we decide to call this function from a different function?

wd_helper.au3
mlipok: Fixed

wd_helper.au3
mlipok: Then we will think how to change them.

wd_helper.au3
mlipok: > 2\. My question was regarding the comma at the beginning of the string, not spaces.

It was desinge this way before:
```AutoIt
		Case 'doubleclick'
			$sPostHoverAction = ',{"button":' & $iButton & ',"type":"pointerDown"},{"button":' & $iButton & ',"type":"pointerUp"},{"button":' & $iButton & ',"type":"pointerDown"},{"button":' & $iButton & ',"type":"pointerUp"}'

		Case 'rightclick'
			$sPostHoverAction = ',{"button":2,"type":"pointerDown"},{"button":2,"type":"pointerUp"}'

		Case 'clickandhold'
			$sPostHoverAction = ',{"button":' & $iButton & ',"type":"pointerDown"},{"type": "pause", "duration": ' & $iHoldDelay & '},{"button":' & $iButton & ',"type":"pointerUp"}'
```

mlipok: It is because of easier way to ADD it to the "TEMPLATE JSON" 

```AutoIt
			; Build dynamic action string
			$sAction = '{"actions":['

			If $sPreAction Then
				$sAction &= $sPreAction
			EndIf

```

mlipok: > 1. Spaces outside of quotes are optional in JSON, so it's purely an issue of aesthetics
Already removed

wd_helper.au3
mlipok: Fixed

wd_helper.au3
Danp2: I think you are saying to leave it as-is. If so, then I disagree. 

IMO, it would be better to provide a proper description of the parameter instead of referencing the calling function.

mlipok: So maybe something like this:
```AutoIt
; Parameters ....: $iButton             - an integer value 0 is .... 1 ... 2 ....
;                  $sAction             - a string value - action type that should be performed
```

mlipok: btw. what about 
`; Link ..........:`
Do you have any link to reference documentation ?
I'm courious how you find out what `$iButton` means and what kind of $sAction can be taken.

Danp2: This [link](https://www.w3.org/TR/webdriver/#actions) is already present in header of _WD_Action, so I don't see the need to add it elsewhere.

wd_helper.au3
Danp2: > It was designed this way before:

Agreed. I was simply asking if it would make sense to make the function more flexible by having the comma be optional. For example --

```autoit
Func __WD_JsonButtonAction($iButton, $sAction, $sPreText = Default, $sPostText = Default)

	If $sPreText = Default Then $sPreText = ','
	If $sPostText = Default Then $sPostText = ''

	Return _
			$sPreText & _
			'{' & _
			'	"button": ' & $iButton & _
			'	,"type": "' & $sAction & '"' & _
			'}' & _
			$sPostText
EndFunc   ;==>__WD_JsonButtonAction
 ```

Ultimately, this would only be useful if __WD_JsonButtonAction is called from additional functions that may not need or want the leading comma.

File: wd_helper.au3 (hunk 4 above)
Comment (mlipok) [reply wd_helper.au3]: Fixed:
https://github.com/mlipok/WebDriver/commit/3d3c9454ad9f6f446c2c16deb7e4a23d57bbbaab

File: wd_helper.au3 (hunk 1 above)
Comment (mlipok) [reply wd_helper.au3]: I will answer this weekend

Comment (mlipok) [reply wd_helper.au3]: as so far `__WD_JsonButtonAction` is used here:

```AutoIt
		Case 'doubleclick'
			$sPostHoverAction = _
					__WD_JsonButtonAction($iButton, "pointerDown") & _
					__WD_JsonButtonAction($iButton, "pointerUp") & _
					__WD_JsonButtonAction($iButton, "pointerDown") & _
					__WD_JsonButtonAction($iButton, "pointerUp")

		Case 'rightclick'
			$sPostHoverAction = _
					__WD_JsonButtonAction(2, "pointerDown") & _
					__WD_JsonButtonAction(2, "pointerUp")

		Case 'clickandhold'
			$sPostHoverAction = _
					__WD_JsonButtonAction($iButton, "pointerDown") & _
					', {' & _
					'	"type":"pause"' & _
					'	,"duration":' & $iHoldDelay & _
					'}' & _
					__WD_JsonButtonAction($iButton, "pointerUp")
....................
....................
....................

		Case 'modifierclick'
			; Hold modifier key down
			$sPreAction = _
					'{' & _
					'	"type":"key"' & _
					'	,"id":"keyboard_1"' & _
					'	,"actions":[' & _
					'		{' & _
					'			"type":"keyDown"' & _
					'			,"value":"' & $sModifier & _
					'		}' & _
					'	]' & _
					'},'

			; Perform click
			$sPostHoverAction = _
					__WD_JsonButtonAction($iButton, "pointerDown") & _
					__WD_JsonButtonAction($iButton, "pointerUp")

```

this mean that `__WD_JsonButtonAction`  only affects `$sPostHoverAction`

which is used only here:

```AutoIt
	Local Static $sActionTemplate = StringReplace( _
			'{' & _
			'	"actions":[' & _ ; Open main action
			'		%s' & _ ; %s > $sPreAction
			'		{' & _ ; Start of default "hover" action
			'			"id":"hover"' & _
			'			,"type":"pointer"' & _
			'			,"parameters":{"pointerType":"mouse"}' & _
			'			,"actions":[' & _ ; Open mouse actions
			'				{' & _
			'					"type":"pointerMove"' & _
			'					,"duration":100' & _
			'					,"x":%s' & _ ; %s > $iXOffset
			'					,"y":%s' & _ ; %s > $iYOffset
			'					,"origin":{' & _
			'						"ELEMENT":"%s"' & _ ; %s > $sElement
			'						,"' & $_WD_ELEMENT_ID & '":"%s"' & _ ; %s > $sElement
			'					}' & _
			'				}' & _
			'				%s' & _ ; %s > $sPostHoverAction
			'			]' & _ ; Close mouse actions
			'		}' & _ ; End of default 'hover' action
			'		%s' & _ ; %s > $sPostAction
			'	]' & _ ; Close main action
			'}', @TAB, '')

```
specifically here:

```AutoIt
			'				}' & _
			'				%s' & _ ; %s > $sPostHoverAction
			'			]' & _ ; Close mouse actions

```

Why to you want to complicate it ?

mlipok: Is it just because you don't like the comma at the beginning of the line?

mlipok: In my experience, this is impractical, I know it is more standard-compliant, but not practical.
Therefore, also when creating SQL queries, I put a comma at the beginning of the line.

I've had many cases where when modifying a SQL query or JSON string, I forgot to control "commas" at the end of a line, but since I started using "commas" at the beginning of a line, all problems are a thing of the past.

And this very case shows that it's easier to add commas in front.

Danp2: My most recent thinking is to change `__WD_JsonButtonAction` (an internal function) into `_WD_JsonAction`, which could also be called by external scripts. This is what I have thus far --
```autoit
Func _WD_JsonAction($sAction, $iValue, $sType = "", $sKey = "")
Local $sJSON = ''

	Switch $sAction
		Case 'mouse'
			$sJSON = _
					'{' & _
					'	"button":' & $iValue & _
					'	,"type":"' & $sType & '"' & _
					'}'
		Case 'pause'
			$sJSON = _
					'{' & _
					'	"type":"pause"' & _
					'	,"duration":' & $iValue & _
					'}'

		Case 'key'
			$sJSON = _
					'{' & _
					'	"type":"key"' & _
					'	,"id":"keyboard_' & $iValue & '"' & _
					'	,"actions":[' & _
					'		{' & _
					'			"type":"' & $sType & '"' & _
					'			,"value":"' & $sKey & _
					'		}' & _
					'	]' & _
					'}'

	EndSwitch

	Return $sJSON
EndFunc   ;==>__WD_JsonAction
```

This allows the calling routine to build a proper JSON "action" without concern for the underlying formatting. It doesn't currently insert the leading comma, so that would need to be handled by the calling routine unless I add that as an option.

Your thoughts on this concept?

wd_helper.au3
```diff
@@ -1859,8 +1859,8 @@ Func _WD_ElementActionEx($sSession, $sElement, $sCommand, $iXOffset = Default, $
 
 	EndSwitch
 
-	; This line in compilation process will be linearized, and will be processed once, thus next usage will be significantly faster
-	; $sActionTemplate declaration is outside the switch to not pollute simplicity of the >Switch ... EndSwitch< - for better code maintenance
+	; $sActionTemplate declaration is outside the switch to not pollute simplicity of the >Switch ... EndSwitch< - for better code maintaince
+	; This following line in compilation process will be linearized, and will be processed once, thus next usage will be significantly faster
```
Danp2, line 1863: I don't believe this is needed or helpful to the reader because --

- "This line" implicitly references the next line of code IMO, so the word "following" isn't needed
- "this following" isn't proper English. "The following" is better syntax *if* you don't agree with the above
- My prior commit fixed a spelling error, which you reintroduce here

wd_helper.au3
```diff
@@ -1889,9 +1889,12 @@ Func _WD_ElementActionEx($sSession, $sElement, $sCommand, $iXOffset = Default, $
 
 	Switch $iActionType
 		Case 1
-			$sAction = StringFormat($sActionTemplate, $sPreAction, $iXOffset, $iYOffset, $sElement, $sElement, $sPostHoverAction, $sPostAction) ; StringFormat() usage is significantly faster than building JSON string each time from scratch
+			; StringFormat() usage is significantly faster than building JSON string each time from scratch
+			; StringReplace() removes all possible @TAB's because they was used only for indentation and are not needed in JSON string
+			$sAction = StringReplace(StringFormat($sActionTemplate, $sPreAction, $iXOffset, $iYOffset, $sElement, $sElement, $sPostHoverAction, $sPostAction), @TAB, '')
```
Danp2, line 1894: - Would the StringFormat comment be placed with the other comments above where we define sActionTemplate?
- Is this StringReplace necessary? I thought this was already handled when the static variable is declared?

Danp2 (changes requested):

wd_helper.au3
Danp2: > Why to you want to complicate it ?

That isn't my intent. 

> Is it just because you don't like the comma at the beginning of the line?

It's not that. My goal would be to make it so that this code could be used in more than one routine.

mlipok: > Your thoughts on this concept?

checking

mlipok: Please take a look now

TheDcoder:
> @Danp2 , @TheDcoder and @seadoggie01 please check all this JSON string reformating provided in this PR.

I am not familiar with the Action API so I don't think I am in a position to judge the code.

I would however like to mention the possibility of using JSON UDF to generate the string for actions instead of hardcoding it and using `StringFormat` on it. One potential issue with the current approach is that it doesn't escape values, which may or may not be fatal in our use case.

That's all the feedback I have for now, best of luck!

Danp2:
@TheDcoder Thanks for the feedback. I've had the same thought about using the JSON functions to build the string, but it's not something that I've pursued thus far.

wd_helper.au3
mlipok: Thanks really for the English lesson. As you know, I am self-taught in this regard.
Will fix it.

wd_helper.au3
mlipok: > * Is this StringReplace necessary? I thought this was already handled when the static variable is declared?

It strips all the possible Tab's from `__WD_JsonAction`

mlipok: > * Would the StringFormat comment be better placed with the other comments above where we define sActionTemplate?

I would prefer to leave this comment in the place where she is. 
Especailly if we remove `StringReplace` from static, and also when we leave the comment about `StringReplace`

wd_helper.au3
```diff
@@ -2098,9 +2098,9 @@ Func __WD_JsonElement($sElement)
 EndFunc   ;==>__WD_JsonElement
 
 ; #FUNCTION# ====================================================================================================================
-; Name ..........: __WD_JsonAction
+; Name ..........: _WD_JsonAction
```
mlipok, line 2101: @Danp2
Why did you change __WD_JsonAction >> _WD_JsonAction
I was thinking about this function as an internal .

wd_helper.au3
```diff
@@ -2098,9 +2098,9 @@ Func __WD_JsonElement($sElement)
 EndFunc   ;==>__WD_JsonElement
 
 ; #FUNCTION# ====================================================================================================================
```
mlipok, line 2100: Also could I ask you about filling out this function header.
As this was your concept and I'm not so good in English.

wd_helper.au3
mlipok: > * Would the StringFormat comment be better placed with the other comments above where we define sActionTemplate?

On the other hand why not.
Especially to not pollute simplicity of the >Switch ... EndSwitch<
;)

mlipok:
> One potential issue with the current approach is that it doesn't escape values, which may or may not be fatal in our use case.

Do not understand `it doesn't escape values`.
Please elaborate or show examples.

mlipok:
> I would however like to mention the possibility of using JSON UDF to generate the string for actions instead of hardcoding it and using `StringFormat` on it.

My aproach is that `_WD_ElementActionEx` will be used in huge loops, and will need to be fast. 
IMHO when we use JSON UDF will slow down this `_WD_ElementActionEx` .

wd_helper.au3
Danp2: > We can remove StringReplace from Static declaration.

What's wrong with leaving it like it was originally written (in the Static declaration)?

wd_helper.au3
Danp2: > Why did you change __WD_JsonAction >> _WD_JsonAction I was thinking about this function as an internal .

@mlipok I did this because I specifically stated that this function would be able to be called by an external routine, and it can be used by the script writer to more easily build webdriver actions.

Danp2:
> Do not understand it doesn't escape values.

https://www.tutorialspoint.com/json_simple/json_simple_escape_characters.htm

We have existing function __WD_EscapeString that can be used for this if needed.

Danp2 (approved):

wd_helper.au3
mlipok: Ok

PR #139, opened 2022-01-20
added _WD_UpdateDriver() to wd_demo.au3

mlipok:


wd_demo.au3
```diff
@@ -458,6 +458,7 @@ Func _USER_WD_Sleep($iDelay)
 EndFunc   ;==>_USER_WD_Sleep
 
 Func SetupGecko()
+	_WD_UpdateDriver('firefox', @ScriptDir , False)
```
Danp2, line 461: Not sue why you specify False for the 3rd parameter when this is the default value.

Danp2 (changes requested):
I would prefer that this be implemented differently --

- No "blind" updating
- Code doesn't need to be in the setup routine
- Options could be added to the GUI to allow user to check for updated browser, then perform update if they choose
- Allow user to choose 32/64 bit

wd_demo.au3
mlipok: ok, will support all yours recomendation which you post in above comments.

mlipok:
> * No "blind" updating

Tell me what you meant here.

Danp2:
> No "blind" updating

I mean don't perform the update automatically without user request / approval. Your initial implementation always downloaded the 32bit version, which may not be what the user wants and could overwrite the 64bit version that they previously installed.

Expanding the GUI to allow the user to control these settings should resolve this concern.

mlipok:
Is it good direction ?

wd_demo.au3
```diff
@@ -51,34 +51,50 @@ Exit
...
 Func _WD_Demo()
 	Local $nMsg
-	Local $iSpacing = 50
+	Local $iPos
+	Local $iSpacing = 25
 	Local $iCount = UBound($aDemoSuite)
 	Local $aCheckboxes[$iCount]
 
-	Local $hGUI = GUICreate("Webdriver Demo", 200, 150 + (20 * $iCount), 100, 200, BitXOR($GUI_SS_DEFAULT_GUI, $WS_MINIMIZEBOX))
+	Local $hGUI = GUICreate("Webdriver Demo", 200, 100, 100, 200, BitXOR($GUI_SS_DEFAULT_GUI, $WS_MINIMIZEBOX))
```
Danp2, line 59: This was coded so that the GUI would dynamically expand as additional demo routines were added. Why would you want to remove this "feature"?

Danp2:
> Is it good direction ?

Yes, it seems good to me. A couple of additional options that could be implemented --

- Report Only - Checks for update, but doesn't perform update if one is available
- Maintain - Detects "bitness" (32 or 64) of existing driver and updates to same 

I definitely want the first one in there. The other one is up to you. 👍

wd_demo.au3
mlipok: focus on:
```AutoIt
	; Resize window
	WinMove($hGUI, "", 100, 200, 200, $iPos + 3 * $iSpacing)
```

I think this is a better way to achieve this what you are talking about.

mlipok:
> * Report Only - Checks for update, but doesn't perform update if one is available

added "Null - check"

mlipok:
> * Maintain - Detects "bitness" (32 or 64) of existing driver and updates to same

I have some idea, need some time to think about.

wd_demo.au3
```diff
@@ -75,8 +75,8 @@ Func _WD_Demo()
 
 	$iPos += $iSpacing
 	GUICtrlCreateLabel("Update", 15, $iPos + 2)
-	Local $idUpdate = GUICtrlCreateCombo("No", 75, $iPos, 100, 20, $CBS_DROPDOWNLIST)
-	GUICtrlSetData($idUpdate, "32bit|32bit+Force|64Bit|64Bit+Force", "No")
+	Local $idUpdate = GUICtrlCreateCombo("Null - check", 75, $iPos, 100, 20, $CBS_DROPDOWNLIST)
```
Danp2, line 83: Not a great description IMO. Would prefer something like "Report only" or "Check; No update".

wd_demo.au3
mlipok: Resolved ?

Danp2 (approved):

Danp2:
> > * Maintain - Detects "bitness" (32 or 64) of existing driver and updates to same
> 
> I have some idea, need some time to think about.

@mlipok Ok to merge this PR or do you want me to hold off for the above?

mlipok:
> @mlipok Ok to merge this PR or do you want me to hold off for the above?

give me next few hours.

mlipok:
> * Maintain - Detects "bitness" (32 or 64) of existing driver and updates to same

Detecting and reporting is already included in the recent PR:
https://github.com/Danp2/WebDriver/pull/149

update to exact Bit version is already done here:
```AutoIt
			; Determine current local webdriver Architecture
			If FileExists($sInstallDir & $sDriverEXE) Then
				_WinAPI_GetBinaryType($sInstallDir & $sDriverEXE)
				$bDriverIs64Bit = (@extended = $SCS_64BIT_BINARY)
				If $sBrowser <> 'chrome' And $bDriverIs64Bit <> $bFlag64 Then
					$bForce = True
;~ 					If $WDDebugSave = $_WD_DEBUG_Info Then
;~ 						__WD_ConsoleWrite($sFuncName & ': ' & $sDriverEXE & ' = ' & (($bDriverIs64Bit) ? ("switching 64>32 Bit") : ("switching 32>64 Bit")) & @CRLF)
;~ 					EndIf
				EndIf
			EndIf

```

Danp2:
I think you misunderstood my intentions. When I stated
>  Maintain - Detects "bitness" (32 or 64) of existing driver and updates to same

I intended for this to be an option added to wd_demo, similar to "Report Only" that was recently added. The goal was to demonstrate how to detect the current bit level (32 vs 64) and then pass the correct value to _WD_UpdateDriver

mlipok:
> The goal was to demonstrate how to detect the current bit level (32 vs 64) and then pass the correct value to _WD_UpdateDriver

Aha... Ok I will provide such modification.
btw. the PR mentioned above is still valid as for me.

mlipok:
> Ok I will provide such modification.

I think that's what the UDF should do.
I will show it in one of the next amendments.

mlipok:
> The goal was to demonstrate how to detect the current bit level (32 vs 64) and then pass the correct value to _WD_UpdateDriver

Does this: https://github.com/Danp2/WebDriver/pull/152
fit all your requests from here ?

Danp2:
> Does this: #152 fit all your requests from here ?

Not exactly. The demo still forces you to choose 32/64 bit. Another option could be added that would result in $bFlag64 = Default, so that _WD_UpdateDriver then performs the upgrade to the matching architecture.

mlipok:
> Another option could be added that would result in $bFlag64 = Default, so that _WD_UpdateDriver then performs the upgrade to the matching architecture.

Done. Please review.

PR #142, opened 2022-01-24
_WD_UpdateDriver: Maintain $bForce value when Null

mlipok:
https://github.com/Danp2/WebDriver/issues/141

### DIFF
--- a/wd_helper.au3
+++ b/wd_helper.au3
@@ -1286,7 +1286,7 @@ Func _WD_UpdateDriver($sBrowser, $sInstallDir = Default, $bFlag64 = Default, $bF
 			If FileExists($sInstallDir & $sDriverEXE) Then
 				_WinAPI_GetBinaryType($sInstallDir & $sDriverEXE)
 				Local $bDriverIs64Bit = (@extended = $SCS_64BIT_BINARY)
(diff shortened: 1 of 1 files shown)


wd_helper.au3
```diff
@@ -1353,9 +1354,9 @@ Func _WD_UpdateDriver($sBrowser, $sInstallDir = Default, $bFlag64 = Default, $bF
 			If $iErr = $_WD_ERROR_Success Then
 				Local $bUpdateAvail = (_VersionCompare($sDriverCurrent, $sDriverLatest) < 0) ; 0 - Both versions equal ; 1 - Version1 greater ; -1 - Version2 greater
 
-				; When $bForce parameter equals Null, then return True if newer driver is available
-				If IsKeyword($bForce) = $KEYWORD_NULL And $bUpdateAvail Then
-					$bResult = True
+				If $bNoUpdate Then
+					; Return True if newer driver is available and No Updates flag is set
+					If $bUpdateAvail Then $bResult = True
```
mlipok, line 1359: Would propose:
```AutoIt
$bResult = $bUpdateAvail
; instead
If $bUpdateAvail Then $bResult = True
```

PR #143, opened 2022-01-24
Func DemoAlerts() - Refactoring

mlipok:
btw.
I also notice that `_WD_Alert($sSession, 'sendtext', 'new text')` doesn't work.
Could you confirm this issue ?

### DIFF
--- a/wd_demo.au3
+++ b/wd_demo.au3
@@ -292,6 +292,11 @@ EndFunc   ;==>DemoCookies
 Func DemoAlerts()
 	Local $sStatus, $sText
 
+	; check status before displaying Alert
+	; check status before displaying Alert
+	$sStatus = _WD_Alert($sSession, 'status')
+	ConsoleWrite("- " & 'Alert Detected => ' & $sStatus & @CRLF)
+
 	; show Alert for testing
 	_WD_ExecuteScript($sSession, "alert('testing 123')")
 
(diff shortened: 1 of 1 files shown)


mlipok:
All `_WD_***` functions should not be placed in `ConsoleWrite()`
Because `_WD_***` functions throws they own messages to Console so, `ConsoleWrite()` messages will be affected.

Thus this coding style should be used:
```AutoIt
$sText = _WD_Alert($sSession, 'gettext')
ConsoleWrite("- " & 'Text Detected => ' & $sText & @CRLF)
```

instead this kind of code:
```AutoIt
ConsoleWrite("- " & 'Text Detected => ' & _WD_Alert($sSession, 'gettext') & @CRLF)
```

Danp2:
>  I also notice that _WD_Alert($sSession, 'sendtext', 'new text') doesn't work.

Looks like it is designed to work with `window.prompt` only, so the example needs to be updated.

mlipok:
> Looks like it is designed to work with `window.prompt` only, so the example needs to be updated.


Please provide fix.

> You removed the initial status check that occurred before the Alert was shown. This was intended to be there to demonstrate how _WD_Alert responded when there wasn't a visible alert.

Please provide fix.

wd_demo.au3
```diff
@@ -292,6 +292,11 @@ EndFunc   ;==>DemoCookies
 Func DemoAlerts()
 	Local $sStatus, $sText
 
+	; check status before displaying Alert
+	; check status before displaying Alert
```
mlipok, line 296: duplicated comments ?

Danp2: Fixed

Danp2 (approved):

PR #149, opened 2022-01-25
_WD_UpdateDriver - Maintain - Detects "bitness"

mlipok:
https://github.com/Danp2/WebDriver/pull/139#issuecomment-1020250811


Danp2:
I don't believe these changes are necessary as the function was already working as intended and it wasn't my desire to add additional console output.

PR #152, opened 2022-01-25
_WD_UpdateDriver() - $bFlag64 = Default handling

mlipok:
https://github.com/Danp2/WebDriver/pull/139#issuecomment-1021198135

If `$bFlag64 = Default` then update should perform checking which version is used.


mlipok:
in case you aproove my PR then before commit 
```AutoIt
;                  $bFlag64     - [optional] Install 64bit version? Default is False
```
should be chagned

PR #154, opened 2022-01-26
added DemoPrint() to wd_demo.au3

mlipok:


Danp2 (commented):
I think Firefox will work without the headless option. Also, you could add this to the GUI as a checkbox.

mlipok:
> Also, you could add this to the GUI as a checkbox.

As I said:
I will make a separate option as a separate PR.

mlipok:
Please in first step end your review for:
https://github.com/Danp2/WebDriver/pull/139

Then this `154`

Danp2 (changes requested):
I'm not understanding why this needed to be pushed as a commit. Can't you test this locally instead?

Danp2 (commented):
> Chrome and Edge Drivers works well with numerical values but Gecko fails.

This is a known issue with  geckodriver -- https://bugzilla.mozilla.org/show_bug.cgi?id=1682497

wd_demo.au3
```diff
@@ -128,7 +128,12 @@ Func RunDemo($idDebugging, $idBrowsers)
 		For $iIndex = 0 To UBound($aDemoSuite, $UBOUND_ROWS) - 1
 			If $aDemoSuite[$iIndex][1] Then
 				ConsoleWrite("+Running: " & $aDemoSuite[$iIndex][0] & @CRLF)
-				Call($aDemoSuite[$iIndex][0])
+				Switch $aDemoSuite[$iIndex][0]
+					Case 'DemoPrint'
+						Call($aDemoSuite[$iIndex][0], $idBrowsers)
+					Case Else
+						Call($aDemoSuite[$iIndex][0])
+				EndSwitch
```
Danp2, line 136: We should look for a better way to implement this that doesn't involve hardcoding function names. I think we could make this data driven by adding a column to $aDemoSuite. Also, it would make more sense to me to pass the browser name instead of the control ID. This would prevent the need from having code like `$aBrowsers[_GUICtrlComboBox_GetCurSel($idBrowsers)][0]` in multiple routines.

Danp2 (changes requested):

wd_demo.au3
mlipok: [Ok I made new ISSUE](https://github.com/Danp2/WebDriver/issues/155) as this should be in separate PR

Danp2 (approved):

PR #162, opened 2022-01-27
wd_demo.au3 + DemoAttach() + wd_demoAttach.au3

mlipok:
I'm not sure how exactly attaching should works.

I was trying to guess using this 2 following links:

https://www.autoitscript.com/wiki/WebDriver#FAQ
https://www.autoitscript.com/forum/topic/201106-webdriver-udf-help-support-ii/?do=findComment&comment=1444173

Please revew/change or propose some fixes.

### DIFF
--- a/wd_demo.au3
+++ b/wd_demo.au3
@@ -429,7 +429,7 @@ Func DemoFrames()
 EndFunc   ;==>DemoFrames
 
 Func DemoActions()
-	Local $sElement, $aElements, $sValue, $sAction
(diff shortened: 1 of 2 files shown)


Danp2:
Attaching to a preexisting browser instance is an advanced matter that doesn't belong in this demo. I would recommend sticking to the wiki if you want to address this issue.

P.S. `--connect-existing  --marionette-port 2828'` is only valid with Firefox, so this won't work with the Chromium based browsers.

PR #165, opened 2022-01-28
_WD_ElementSelectAction() - 'options' speed up

mlipok:
The discusion was started here:
https://www.autoitscript.com/forum/topic/191990-webdriver-udf-w3c-compliant-version-01192022/?do=findComment&comment=1494610

### DIFF
--- a/wd_core.au3
+++ b/wd_core.au3
@@ -105,7 +105,6 @@ Global Enum _
 		$_WD_ERROR_SessionInvalid, _ ;
 		$_WD_ERROR_UnknownCommand, _ ;
 		$_WD_ERROR_UserAbort, _ ;
-		$_WD_ERROR_HtmlProcessing, _ ;
 		$_WD_ERROR_COUNTER ;
 
 Global Const $aWD_ERROR_DESC[$_WD_ERROR_COUNTER] = [ _
@@ -126,8 +125,7 @@ Global Const $aWD_ERROR_DESC[$_WD_ERROR_COUNTER] = [ _
 		"Element interaction issue", _
 		"Invalid session ID", _
 		"Unknown Command", _
-		"User Aborted", _
-		"Error processing HTML" _
+		"User Aborted" _
 		]
 
 Global Const $WD_ErrorInvalidSession = "invalid session id"
--- a/wd_core.au3
+++ b/wd_core.au3
@@ -105,6 +105,7 @@ Global Enum _
 		$_WD_ERROR_SessionInvalid, _ ;
 		$_WD_ERROR_UnknownCommand, _ ;
 		$_WD_ERROR_UserAbort, _ ;
+		$_WD_ERROR_HtmlProcessing, _ ;
 		$_WD_ERROR_COUNTER ;
 
 Global Const $aWD_ERROR_DESC[$_WD_ERROR_COUNTER] = [ _
--- a/wd_core.au3
+++ b/wd_core.au3
@@ -126,7 +126,8 @@ Global Const $aWD_ERROR_DESC[$_WD_ERROR_COUNTER] = [ _
 		"Element interaction issue", _
 		"Invalid session ID", _
 		"Unknown Command", _
-		"User Aborted" _
+		"User Aborted", _
+		"Error processing HTML" _
 		]
 
 Global Const $WD_ErrorInvalidSession = "invalid session id"
--- a/wd_helper.au3
+++ b/wd_helper.au3
@@ -976,6 +976,7 @@ EndFunc   ;==>_WD_ElementOptionSelect
 ;                  $sSelectElement - Element ID of Select element from _WD_FindElement
 ;                  $sCommand       - Action to be performed. Can be one of the following:
 ;                  |OPTIONS - Retrieve array containing value / label attributes from the Select element's options
+;                  |OPTIONS2 - Retrieve array containing value / label attributes from the Select element's options - using faster way - RegExp
 ;                  |VALUE - Retrieve current value
 ; Return values .: Success - Requested data returned by web driver.
 ;                  Failure - "" (empty string) and sets @error to one of the following values:
@@ -995,6 +996,7 @@ Func _WD_ElementSelectAction($sSession, $sSelectElement, $sCommand)
 	Local Const $sFuncName = "_WD_ElementSelectAction"
 	Local $sNodeName, $sJsonElement, $vResult
 	Local $sText, $aOptions
+	Local $iSelectedOption = -1
 
 	$sNodeName = _WD_ElementAction($sSession, $sSelectElement, 'property', 'nodeName')
 	Local $iErr = @error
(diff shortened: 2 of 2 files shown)


wd_core.au3
```diff
@@ -105,6 +105,7 @@ Global Enum _
 		$_WD_ERROR_SessionInvalid, _ ;
 		$_WD_ERROR_UnknownCommand, _ ;
 		$_WD_ERROR_UserAbort, _ ;
+		$_WD_ERROR_HtmlProcessing, _ ;
```
Danp2, line 108: You would also need to add a matching entry in $aWD_ERROR_DESC

wd_helper.au3
```diff
@@ -976,6 +976,7 @@ EndFunc   ;==>_WD_ElementOptionSelect
 ;                  $sSelectElement - Element ID of Select element from _WD_FindElement
 ;                  $sCommand       - Action to be performed. Can be one of the following:
 ;                  |OPTIONS - Retrieve array containing value / label attributes from the Select element's options
+;                  |OPTIONS2 - Retrieve array containing value / label attributes from the Select element's options - using faster way - RegExp
```
Danp2, line 979: I suggest that we stick to a single 'Options' command. Either we replace the existing code with your version or perhaps there's a way to improve the existing version so that it doesn't loop through each option with WD commands.

wd_helper.au3
```diff
@@ -1037,7 +1080,12 @@ Func _WD_ElementSelectAction($sSession, $sSelectElement, $sCommand)
 		__WD_ConsoleWrite($sFuncName & ': ' & ((IsArray($vResult)) ? "(array)" : $vResult) & @CRLF)
 	EndIf
 
-	Return SetError(__WD_Error($sFuncName, $iErr), $_WD_HTTPRESULT, $vResult)
+	Local $iExt = $_WD_HTTPRESULT
+
+	; if $iSelectedOption was found then pass the ID ($iSelectedOption) to @extended
+	If $_WD_HTTPRESULT = $HTTP_STATUS_OK And $iSelectedOption <> -1 Then $iExt = $iSelectedOption
```
Danp2, line 1086: Not sure why you are checking $_WD_HTTPRESULT here. Wouldn't it be more appropriate to check $iErr?

Danp2 (changes requested):

wd_core.au3
mlipok: suplemented

wd_helper.au3
mlipok: I agree with @TheDcoder  opinions
https://www.autoitscript.com/forum/topic/191990-webdriver-udf-w3c-compliant-version-01192022/?do=findComment&comment=1494624
`we also need to keep the current version as it is more robust, the RegEx might fail in edge cases, there's a saying about using RegEx with HTML`

wd_helper.au3
mlipok: Done in commit `$iExt = $iSelectedOption`

wd_helper.au3
```diff
@@ -1082,8 +1082,8 @@ Func _WD_ElementSelectAction($sSession, $sSelectElement, $sCommand)
 
 	Local $iExt = $_WD_HTTPRESULT
 
-	; if $iSelectedOption was found then pass the ID ($iSelectedOption) to @extended
-	If $_WD_HTTPRESULT = $HTTP_STATUS_OK And $iSelectedOption <> -1 Then $iExt = $iSelectedOption
+	; if OPTIONS2 was succesfully used then pass the ID ($iSelectedOption) to @extended
```
mlipok, line 1085: Description in Function Header should be added.
@Danp2  can I leave this part of the job to you ?

wd_helper.au3
Danp2: Yes, that makes perfect sense. Thanks for reminding me about that prior discussion

Danp2: I'm still unsure about the "options2" label. Can you come up with something that better describes this new command?

wd_helper.au3
Danp2: The meaning of @extended shouldn't vary within the function. If you want to return the "selected" status, then this should be done for both "options" routines. Have you considered adding another column to the array?

mlipok: For example what type of column ?

Danp2: Boolean is my initial thought

mlipok: For selected ? 
If so I prefer to return them in EXT as you have it direclty an you do not must process them outside the function to check which one is selected.

Danp2: Please keep the following in mind --

- This function wasn't designed to return the element's selected status
- The current "options" implementation could be modified so that it is both reliable and fast

One option, which I actually like, is to modify _WD_ElementSelectAction to support a "selected" command

mlipok: > One option, which I actually like, is to modify _WD_ElementSelectAction to support a "selected" command

Ok I will provide such option soon.

Danp2: You may want to submit it as a separate PR so it doesn't get held back by this one

mlipok: Ok

Danp2: I'm no longer convinced that adding a "selected" command is a good thing. I believe the current "value" command already gives us the selected item's value. So what purpose would "selected" command serve?

wd_helper.au3
```diff
@@ -1025,6 +1027,47 @@ Func _WD_ElementSelectAction($sSession, $sSelectElement, $sCommand)
...
+					Else
+						Local $aTemp
+						
+						; check which element is selected
+						For $iCheck = 0 To UBound($aOptions_outer) - 1
+							$aTemp = $aOptions_outer[$iCheck]
+							If StringInStr($aTemp[3], 'selected="selected"') Then
+								$iSelectedOption = $iCheck
+								ExitLoop
+							EndIf
```
Danp2, line 1051: I don't believe this works the way you intended. AFAICS, `'selected="selected"'` only establishes the initially selected option and doesn't change when a new item is selected.

wd_helper.au3
mlipok: So the descriptions 
```AutoIt
; |VALUE - Retrieve current value
```
could be more adequate, something like
```AutoIt
; |VALUE - Retrieve current selected value
```
or
```AutoIt
; |VALUE - Retrieve current value - selected item
```

mlipok: > So what purpose would "selected" command serve?

None.

mlipok: > Please keep the following in mind --
> 
> * This function wasn't designed to return the element's selected status
> * The current "options" implementation could be modified so that it is both reliable and fast
> 
> One option, which I actually like, is to modify _WD_ElementSelectAction to support a "selected" command

So what we do now with this ?
Do you mean that `OPTIONS2` should not return the selected value in EXTENDED ?

Danp2: We may not need to include an `Options2` if a suitable replacement for the existing `Options` is created. Here's my initial attempt --

```autoit
 Func _WD_ElementSelectAction2($sSession, $sSelectElement, $sCommand)
	Local Const $sFuncName = "_WD_ElementSelectAction2"
	Local $sNodeName, $sJsonElement, $vResult
	Local $sText, $aOptions

	$sNodeName = _WD_ElementAction($sSession, $sSelectElement, 'property', 'nodeName')
	Local $iErr = @error

	If $iErr = $_WD_ERROR_Success And $sNodeName = 'select' Then

		Switch $sCommand
			Case 'value'
				; Retrieve current value of designated Select element
				$sJsonElement = __WD_JsonElement($sSelectElement)
				$vResult = _WD_ExecuteScript($sSession, "return arguments[0].value", $sJsonElement, Default, $_WD_JSON_Value)
				$iErr = @error

			Case 'options'
				; Retrieve array containing value / label attributes from the Select element's options
				Local $sScript = "var result =''; var options = arguments[0].options; for (let i = 0; i < options.length; i++) {result += options[i].value + '|' + options[i].label + '\n'} return result;"
				Local $sText = _WD_ExecuteScript($sSession, $sScript, __WD_JsonElement($sSelect), Default, $_WD_JSON_Value)
				$iErr = @error

				$sText = StringStripWS($sText, $STR_STRIPTRAILING)
				Local $aOut[0][2]
				_ArrayAdd($aOut, $sText, 0, Default, @LF, 1)
				$vResult = $aOut
			Case Else
				Return SetError(__WD_Error($sFuncName, $_WD_ERROR_InvalidDataType, "(Value|Options) $sCommand=>" & $sCommand), 0, "")

		EndSwitch
	Else
	EndIf

		__WD_ConsoleWrite($sFuncName & ': ' & ((IsArray($vResult)) ? "(array)" : $vResult) & @CRLF)
	EndIf

	Return SetError(__WD_Error($sFuncName, $iErr), $_WD_HTTPRESULT, $vResult)
EndFunc   ;==>_WD_ElementSelectAction2
```

Would be interesting to see how the speed compares to your version.

mlipok: > We may not need to include an `Options2` if a suitable replacement for the existing `Options` is created. Here's my initial attempt --

yes

> Would be interesting to see how the speed compares to your version.

Durations = 8.855816 sec : OLD
Durations = 0.041332 sec : HTML PROCESSING
Durations = 0.0332274 sec : JAVA SCRIPT

mlipok: Look at my recent commits.
btw. this conversation could be resolved.

wd_helper.au3
mlipok: This is no longer needed.

wd_helper.au3
mlipok: This is no longer needed.

Danp2 (approved):

PR #169, opened 2022-01-31
__WD_JsonElement() usage

mlipok:


Danp2:
@mlipok These changes appear to be more about your style preference than actually improving or fixing an issue. Am I missing something?

mlipok:
I saw this coding style here

https://github.com/Danp2/WebDriver/pull/165/files/c31293e328b1d21156c53f9249638bc1b4a8311f..674f09a468223fe2e1a2bfb9da0951b2950d3df0

```AutoIt
				Case 'options'
					; Retrieve array containing value / label attributes from the Select element's options
					Local $sScript = "var result =''; var options = arguments[0].options; for (let i = 0; i < options.length; i++) {result += options[i].value + '|' + options[i].label + '\n'} return result;"
					$vResult = _WD_ExecuteScript($sSession, $sScript, __WD_JsonElement($sSelectElement), Default, $_WD_JSON_Value)
					$iErr = @error

```

Danp2:
I don't believe that it is necessary to go back and change working code because of a design decision made on a more recent commit. However, I will go ahead and accpet your PR because the changes are relatively simple and unlikely to introduce any bugs. 😉

PR #172, opened 2022-02-02
_WD_HighlightElements refactoring

Danp2:
- No longer calls _WD_HighlightElement in a loop
- Looping now handled in Javascript

### DIFF
--- a/wd_helper.au3
+++ b/wd_helper.au3
@@ -632,8 +632,8 @@ Func _WD_FrameLeave($sSession)
 EndFunc   ;==>_WD_FrameLeave
 
 ; #FUNCTION# ===========================================================================================================
-; Name ..........: [Deprecated] _WD_HighlightElement
-; Description ...: Highlights the specified element.
+; Name ..........: _WD_HighlightElement
+; Description ...: Highlights the specified element. <B>[Deprecated]</B>
 ; Syntax ........: _WD_HighlightElement($sSession, $sElement[, $iMethod = Default])
 ; Parameters ....: $sSession - Session ID from _WD_CreateSession
 ;                  $sElement - Element ID from _WD_FindElement
@@ -667,7 +667,7 @@ EndFunc   ;==>_WD_HighlightElement
 ; #FUNCTION# ====================================================================================================================
 ; Name ..........: _WD_HighlightElements
 ; Description ...: Highlights the specified elements.
-; Syntax ........: _WD_HighlightElements($sSession, $aElements[, $iMethod = Default])
+; Syntax ........: _WD_HighlightElements($sSession, $vElements[, $iMethod = Default])
 ; Parameters ....: $sSession  - Session ID from _WD_CreateSession
 ;                  $vElements - Element ID from _WD_FindElement (single element as string; multiple elements as array)
 ;                  $iMethod   - [optional] an integer value to set the style (default = 1)
@@ -680,7 +680,7 @@ EndFunc   ;==>_WD_HighlightElement
 ; Author ........: Danyfirex
 ; Modified ......: mLipok, Danp2
 ; Remarks .......:
-; Related .......: _WD_HighlightElement [Deprecated]
+; Related .......:
 ; Link ..........: https://www.autoitscript.com/forum/topic/192730-webdriver-udf-help-support/?do=findComment&comment=1396643
 ; Example .......: No
 ; ===============================================================================================================================
--- a/wd_helper.au3
+++ b/wd_helper.au3
@@ -699,12 +699,11 @@ Func _WD_HighlightElements($sSession, $vElements, $iMethod = Default)
 	If $iMethod < 0 Or $iMethod > 3 Then $iMethod = 1
 
 	If IsString($vElements) Then
-		ConsoleWrite('@@ Debug(' & @ScriptLineNumber & ') : $vElements = ' & $vElements & @CRLF & '>Error code: ' & @error & @CRLF) ;### Debug Console
 		$sScript = "arguments[0].style='" & $aMethod[$iMethod] & "'; return true;"
 		$sResult = _WD_ExecuteScript($sSession, $sScript,  __WD_JsonElement($vElements), Default, $_WD_JSON_Value)
 		$iErr = @error
 
-	ElseIf IsArray($vElements) Then
+	ElseIf IsArray($vElements) And UBound($vElements) > 0 Then
 		For $i = 0 To UBound($vElements) - 1
 			$vElements[$i] = __WD_JsonElement($vElements[$i])
 		Next
--- a/wd_helper.au3
+++ b/wd_helper.au3
@@ -686,8 +686,8 @@ EndFunc   ;==>_WD_HighlightElement
 ;                  1 - Highlight border dotted red
 ;                  2 - Highlight yellow rounded box
 ;                  3 - Highlight yellow rounded box + border  dotted red
-; Return values .: Success - True. @extended is set to the number of highlighted elements
-;                  Failure - False and sets @error to $_WD_ERROR_GeneralError
+; Return values .: Success - True
+;                  Failure - False and sets @error returned from _WD_ExecuteScript()
 ; Author ........: Danyfirex
 ; Modified ......: mLipok
 ; Remarks .......:
@@ -696,14 +696,34 @@ EndFunc   ;==>_WD_HighlightElement
 ; Example .......: No
 ; ===============================================================================================================================
 Func _WD_HighlightElements($sSession, $aElements, $iMethod = Default)
+	Local Const $sFuncName = "_WD_HighlightElements"
+	Local Const $aMethod[] = _
+			[ _
+			"border: 0px", _
+			"border: 2px dotted red", _
+			"background: #FFFF66; border-radius: 5px; padding-left: 3px;", _
+			"border: 2px dotted red; background: #FFFF66; border-radius: 5px; padding-left: 3px;" _
+			]
+
 	Local $iHighlightedElements = 0
 
 	If $iMethod = Default Then $iMethod = 1
+	If $iMethod < 0 Or $iMethod > 3 Then $iMethod = 1
 
 	For $i = 0 To UBound($aElements) - 1
-		$iHighlightedElements += (_WD_HighlightElement($sSession, $aElements[$i], $iMethod) = True ? 1 : 0)
+		$aElements[$i] = __WD_JsonElement($aElements[$i])
 	Next
-	Return ($iHighlightedElements > 0 ? SetError(0, $iHighlightedElements, True) : SetError($_WD_ERROR_GeneralError, 0, False))
+
+	Local $sElements = "[" & _ArrayToString($aElements, ",") & "]"
+	Local $sScript = "for (var i = 0, max = arguments[0].length; i < max; i++) { arguments[0][i].style = '" & $aMethod[$iMethod] & "'; }; return true;"
+	Local $sResult = _WD_ExecuteScript($sSession, $sScript, $sElements, Default, $_WD_JSON_Value)
+	Local $iErr = @error
+
+	If $_WD_DEBUG = $_WD_DEBUG_Info Then
+		__WD_ConsoleWrite($sFuncName & ': ' & $sResult & @CRLF)
+	EndIf
+
+	Return SetError($iErr, $_WD_HTTPRESULT, ($iErr= $_WD_ERROR_Success))
 EndFunc   ;==>_WD_HighlightElements
 
 ; #FUNCTION# ====================================================================================================================
--- a/wd_helper.au3
+++ b/wd_helper.au3
@@ -652,27 +652,16 @@ EndFunc   ;==>_WD_FrameLeave
 ; Example .......: No
 ; ===============================================================================================================================
 Func _WD_HighlightElement($sSession, $sElement, $iMethod = Default)
-	Local Const $sFuncName = "_WD_HighlightElement"
-	Local Const $aMethod[] = _
-			[ _
-			"border: 0px", _
-			"border: 2px dotted red", _
-			"background: #FFFF66; border-radius: 5px; padding-left: 3px;", _
-			"border: 2px dotted red; background: #FFFF66; border-radius: 5px; padding-left: 3px;" _
-			]
+ 	Local Const $sFuncName = "_WD_HighlightElement"
 
-	If $iMethod = Default Then $iMethod = 1
-	If $iMethod < 0 Or $iMethod > 3 Then $iMethod = 1
-
-	Local $sScript = "arguments[0].style='" & $aMethod[$iMethod] & "'; return true;"
-	Local $sResult = _WD_ExecuteScript($sSession, $sScript,  __WD_JsonElement($sElement), Default, $_WD_JSON_Value)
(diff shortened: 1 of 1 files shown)


Danp2:
@mlipok @Sven-Seyfert Please review and let me know if you encounter any problems or identify other needed changes.

wd_helper.au3
```diff
@@ -686,8 +686,8 @@ EndFunc   ;==>_WD_HighlightElement
 ;                  1 - Highlight border dotted red
 ;                  2 - Highlight yellow rounded box
 ;                  3 - Highlight yellow rounded box + border  dotted red
```
Sven-Seyfert, line 688: One whitespace is enough 😅 . Same for the function above `_WD_HighlightElement`.

wd_helper.au3
```diff
@@ -696,14 +696,34 @@ EndFunc   ;==>_WD_HighlightElement
 ; Example .......: No
 ; ===============================================================================================================================
 Func _WD_HighlightElements($sSession, $aElements, $iMethod = Default)
+	Local Const $sFuncName = "_WD_HighlightElements"
+	Local Const $aMethod[] = _
+			[ _
+			"border: 0px", _
+			"border: 2px dotted red", _
+			"background: #FFFF66; border-radius: 5px; padding-left: 3px;", _
+			"border: 2px dotted red; background: #FFFF66; border-radius: 5px; padding-left: 3px;" _
```
Sven-Seyfert, line 705: The first two styles doesn't include the trailing semicolon `;`.
The others do include the trailing semicolon `;`.

I believe it doesn't matter, but I would go with the semicolon for all styles, even when it's just one style.
Same for the function above _WD_HighlightElement.

wd_helper.au3
```diff
@@ -696,14 +696,34 @@ EndFunc   ;==>_WD_HighlightElement
 ; Example .......: No
 ; ===============================================================================================================================
 Func _WD_HighlightElements($sSession, $aElements, $iMethod = Default)
+	Local Const $sFuncName = "_WD_HighlightElements"
+	Local Const $aMethod[] = _
```
mlipok, line 700: The same `$aMethod[]` is used in `_WD_HighlightElement()` maybe should we put this to Globoal  scope ?
Or even can we combine this 2 separate function in one and only check if `IsString($vElement)` or `IsArray($vElement)`  ?
All others looks fine and should work properly.

wd_helper.au3
```diff
@@ -696,14 +696,34 @@ EndFunc   ;==>_WD_HighlightElement
...
+	Local $sElements = "[" & _ArrayToString($aElements, ",") & "]"
+	Local $sScript = "for (var i = 0, max = arguments[0].length; i < max; i++) { arguments[0][i].style = '" & $aMethod[$iMethod] & "'; }; return true;"
+	Local $sResult = _WD_ExecuteScript($sSession, $sScript, $sElements, Default, $_WD_JSON_Value)
+	Local $iErr = @error
+
+	If $_WD_DEBUG = $_WD_DEBUG_Info Then
+		__WD_ConsoleWrite($sFuncName & ': ' & $sResult & @CRLF)
+	EndIf
+
+	Return SetError($iErr, $_WD_HTTPRESULT, ($iErr= $_WD_ERROR_Success))
```
Sven-Seyfert, line 726: I don't know how precise I should be before you think this is unimportant, annoying or just doesn't matter, but let's try it out 😁 :

- In function `_WD_HighlightElement` is a whitespace between `$iErr =`, but not in this function.
- Another difference is `.style='"` which looks a bit different here `.style = '"`.

Sven-Seyfert (approved):
Besides my notes which doesn't really matter, it looks good to me 👍 .

wd_helper.au3
Danp2: I was thinking the same thing about combining the functions.

Sven-Seyfert: Hi @mlipok ,

my tendece is to agree about merging the functions together, because the differences aren't much.
This isn't usually my approach, but at this point could be okay 🤞 .

Best regards
Sven

wd_helper.au3
Danp2: Not sure why it appears like this. They both appear the same on my end, and I didn't change anything in this area AFAIK

wd_helper.au3
Danp2: 👎 J/K 😛

wd_helper.au3
Danp2: Check out the latest commit where I've essentially combined the functionality into a single function. `_WD_HighlightElement` still exists, but it just passes the parameters on to `_WD_HighlightElements`.

Under this scenario, `_WD_HighlightElement` could be marked as deprecated and it would eventually be removed.

wd_helper.au3
```diff
@@ -687,43 +676,52 @@ EndFunc   ;==>_WD_HighlightElement
...
+	Local $sScript, $sResult, $iErr, $sElements
 
 	If $iMethod = Default Then $iMethod = 1
 	If $iMethod < 0 Or $iMethod > 3 Then $iMethod = 1
 
-	For $i = 0 To UBound($aElements) - 1
-		$aElements[$i] = __WD_JsonElement($aElements[$i])
-	Next
+	If IsString($vElements) Then
+		ConsoleWrite('@@ Debug(' & @ScriptLineNumber & ') : $vElements = ' & $vElements & @CRLF & '>Error code: ' & @error & @CRLF) ;### Debug Console
```
mlipok, line 702: ConsoleWrite ?
Why not   __WD_ConsoleWrite  ?

wd_helper.au3
```diff
@@ -687,43 +676,52 @@ EndFunc   ;==>_WD_HighlightElement
...
+		$sScript = "arguments[0].style='" & $aMethod[$iMethod] & "'; return true;"
+		$sResult = _WD_ExecuteScript($sSession, $sScript,  __WD_JsonElement($vElements), Default, $_WD_JSON_Value)
+		$iErr = @error
 
-	Local $sElements = "[" & _ArrayToString($aElements, ",") & "]"
-	Local $sScript = "for (var i = 0, max = arguments[0].length; i < max; i++) { arguments[0][i].style = '" & $aMethod[$iMethod] & "'; }; return true;"
-	Local $sResult = _WD_ExecuteScript($sSession, $sScript, $sElements, Default, $_WD_JSON_Value)
-	Local $iErr = @error
+	ElseIf IsArray($vElements) Then
```
mlipok, line 707: `UBound($vElements)` should be used to be sure any element is given.

mlipok (changes requested):

wd_helper.au3
Danp2: That shouldn't even be there. It's left over from my testing. 🙄

wd_helper.au3
```diff
@@ -652,27 +652,16 @@ EndFunc   ;==>_WD_FrameLeave
...
-	If $iMethod = Default Then $iMethod = 1
-	If $iMethod < 0 Or $iMethod > 3 Then $iMethod = 1
+ 	Local Const $sFuncName = "_WD_HighlightElement"
 
-	Local $sScript = "arguments[0].style='" & $aMethod[$iMethod] & "'; return true;"
-	Local $sResult = _WD_ExecuteScript($sSession, $sScript,  __WD_JsonElement($sElement), Default, $_WD_JSON_Value)
+	Local $bResult = _WD_HighlightElements($sSession, $sElement, $iMethod)
 	Local $iErr = @error
 
 	If $_WD_DEBUG = $_WD_DEBUG_Info Then
```
mlipok, line 660: this is not needed as it is already done in `_WD_HighlightElements`
Thus 
```AutoIt
Local $iErr = @error 
```
is not needed just use:

```AutoIt
Return SetError($iErr, @error, $bResult)
```

mlipok (changes requested):

wd_helper.au3
```diff
@@ -652,27 +652,16 @@ EndFunc   ;==>_WD_FrameLeave
 ; Example .......: No
 ; ===============================================================================================================================
 Func _WD_HighlightElement($sSession, $sElement, $iMethod = Default)
```
mlipok, line 654: I know that this will be `Script Breaking Changed` but 

`_WD_HighlightElement` have exactly the same parameters like `_WD_HighlightElements` and IMO `_WD_HighlightElement` could be deleted.

For script writers it will be very easy to fix. just by changing `_WD_HighlightElement` into `_WD_HighlightElements` string in all scripts for example using `Grep4Win` or manually using SciTE like editors.

wd_helper.au3
Danp2: I don't think so, unless I've misunderstood what you are saying. You have to capture the @error coming out of `_WD_HighlightElements` and then pass it on to the calling routine.

wd_helper.au3
Danp2: That's correct. But I think the correct way to handle this is to mark the function as deprecated and completely remove it later on.

wd_helper.au3
```diff
@@ -699,12 +699,11 @@ Func _WD_HighlightElements($sSession, $vElements, $iMethod = Default)
 	If $iMethod < 0 Or $iMethod > 3 Then $iMethod = 1
 
 	If IsString($vElements) Then
-		ConsoleWrite('@@ Debug(' & @ScriptLineNumber & ') : $vElements = ' & $vElements & @CRLF & '>Error code: ' & @error & @CRLF) ;### Debug Console
 		$sScript = "arguments[0].style='" & $aMethod[$iMethod] & "'; return true;"
 		$sResult = _WD_ExecuteScript($sSession, $sScript,  __WD_JsonElement($vElements), Default, $_WD_JSON_Value)
 		$iErr = @error
 
-	ElseIf IsArray($vElements) Then
+	ElseIf IsArray($vElements) And UBound($vElements) > 0 Then
```
mlipok, line 706: `IsArray($vElements) And` is not needed.
`UBound($vElements)` do the same at once.

mlipok (changes requested):

wd_helper.au3
mlipok: > But I think the correct way to handle this is to mark the function as deprecated and completely remove it later on.

Good idea. Will you set a deletion deadline to 2023?

wd_helper.au3
mlipok: Just 
```AutoIt
ElseIf UBound($vElements) Then
```
It's enough/

mlipok: or... you are preparing script for `map` ?
As `UBound()` can be used also for `map` thus yours `IsArray()` usage ?

wd_helper.au3
mlipok: > You have to capture the @error coming out of `_WD_HighlightElements` and then pass it on to the calling routine.

Almost the same error will be notified firstly by `_WD_HighlightElements` and then again by `_WD_HighlightElement` so they will be duplicated/redundant.

wd_helper.au3
Danp2: Yes, having `IsArray()` in the code will ensure that only variables of the correct type are processed in this section of code.

wd_helper.au3
mlipok: Especially since `_WD_HighlightElement` is just a wrapper for` _WD_HighlightElements`, and also because `_WD_HighlightElement` will be marked as Deprecated.

wd_helper.au3
Danp2: Haven't decided when that will occur. However, I don't think it needs to be a year out.

wd_helper.au3
mlipok: I keep forgetting about maps.
So this part is ok.

wd_helper.au3
Danp2: Clearly we have a misunderstanding. Here are my observations --

- You can't eliminate `Local $iErr = @error` in _WD_HighlightElement without also changing the function's Return line. Otherwise, you get `warning: $iErr: possibly used before declaration.`

- The error code that is generated in `_WD_HighlightElements` needs to "bubble up" through `_WD_HighlightElement` so that the routine that initially called `_WD_HighlightElement` can check the value of @error to determine if an error occurred.

I understand that there will be some duplication as far as any logging to the console, but that isn't a concern for me.
Does that help? If not, what part of this doesn't make sense to you? Be specific. 😃

mlipok: To be specific, this is depracted function which is simple wrapper.
So I propose to truncate it as much as possible.:
```AutoIt
Func _WD_HighlightElement($sSession, $sElement, $iMethod = Default)
	Local $bResult = _WD_HighlightElements($sSession, $sElement, $iMethod)
	Return SetError(@error, @extended, $bResult)
EndFunc   ;==>_WD_HighlightElement
```

Danp2: I'm satisfied with the current revision. It's really a moot point since the function will eventually be removed.

mlipok: Ok.

wd_helper.au3
```diff
@@ -632,7 +632,7 @@ Func _WD_FrameLeave($sSession)
 EndFunc   ;==>_WD_FrameLeave
 
 ; #FUNCTION# ===========================================================================================================
-; Name ..........: _WD_HighlightElement
+; Name ..........: [Deprecated] _WD_HighlightElement
```
mlipok, line 635: `[Deprecated]`
Why in `Name` and not in `Remarks` ?

wd_helper.au3
```diff
@@ -680,7 +680,7 @@ EndFunc   ;==>_WD_HighlightElement
 ; Author ........: Danyfirex
 ; Modified ......: mLipok, Danp2
 ; Remarks .......:
-; Related .......: _WD_HighlightElement
+; Related .......: _WD_HighlightElement [Deprecated]
```
mlipok, line 683: It is uncommon to place here [Deprecated] .
It is even unusual.

UDF Specs says: [A comma seperated list of functions or structures related to the function.](https://www.autoitscript.com/wiki/UDF-spec#Headers)

wd_helper.au3
Danp2: Because I wanted it to be highly visible. I also added remarks that attempt to address the situation.

I will have to check how this shows in the help file, and may have to make further adjustments.

mlipok (changes requested):

wd_helper.au3
Danp2: > It is uncommon to place here [Deprecated] .
It is even unusual.

> UDF Specs says:

- You appear to have issues with my use of "[Deprecated]" in the header. Do you have a suggested alternative?
- The specs are silent AFAICS regarding additional tags like this. Therefore, I would interpret that it is acceptable.
- Who died and named you the spec police? 😜 

Seriously, this `related` entry will be eventually removed when `_WD_HighlightElement` is removed for good so I don't see that it's a big deal. I still need to check the help file functionality with this coding in place, and I may have to make adjustments due to broken links.

mlipok: Ok.
btw.
We just discuss.

Danp2: > btw. We just discuss.

Understood and agree. However, I think it's easy to misinterpret some of your responses. It may be due to the language barrier. It also could just be differences in our communication styles. 🙂

mlipok: Each of them.

PR #186, opened 2022-02-08
_WD_GetSource - Return values suplemented

mlipok:


wd_core.au3
```diff
@@ -945,7 +945,7 @@ EndFunc   ;==>_WD_Alert
 ; Description ...: Get page source.
 ; Syntax ........: _WD_GetSource($sSession)
 ; Parameters ....: $sSession - Session ID from _WD_CreateSession
-; Return values .: Success - Source code from page.
+; Return values .: Success - Source HTML code from page as string serialization of the DOM of the current browsing context active document.
```
Danp2, line 948: Whoa.... that's a mouth full. 😮 Appears to come from the W3C specs, but I don't believe we need this here since the function returns the contents of the `Value` node and not the full serialized string.

Wouldn't this suffice?

`; Return values .: Success - HTML source code from page.`

mlipok: It was related to todays user asking about how to get HTML source

Danp2: Ok. Let me know if you plan to revise that line or just close the PR.

mlipok: Will change as you sugessted.
Give me some time. 
Far from computer.

Danp2 (approved):

PR #194, opened 2022-02-12
_WD_JsonCookie() added

mlipok:
Here is my new proposal


mlipok:
I'm wondering how mulptiple coockies should be added ?

Simply like in this following example:
```AutoIt
	_WD_Cookies($sSession, 'add', _WD_JsonCookie("name1", "value1"))
	_WD_Cookies($sSession, 'add', _WD_JsonCookie("name2", "value2"))
	_WD_Cookies($sSession, 'add', _WD_JsonCookie("name3", "value3"))
```

or for all this 3 different cookies `should or even must` be created single JSON string like this:
```json
 {
   "cookie": [
     {"name": "cookie1", "value": "hello"},
     {"name": "cookie2", "value": "goodbye"}
   ]
 }
```
?

Danp2:
> I'm wondering how mulptiple cookies should be added ?

I haven't seen anywhere that suggests that it could be done in a single command, but this is something you would need to test. 

> also this all feature should be added

Don't forget `Cookie path` and `Cookie domain`

mlipok:
> I haven't seen anywhere that suggests that it could be done in a single command, but this is something you would need to test.

https://github.com/jlipps/simple-wd-spec/
`The Add Cookie command adds a single cookie to the cookie store associated with the active document's address.`

Is it sufficient to confirm how it works?

mlipok:
As for now in `DemoCookies()` i used:
```autoit
Local $sCookie = _WD_JsonCookie($sName, $sValue, Default, 'www.google.com', True, False, 1707825601, "None")
```

When it comes to `$iExpiryTime = 1707825601` I would like to do something like `current time + 1 minute`.

I search for some UNIX EPOCH time on the forum.

I found some discusions/examples here:
https://www.autoitscript.com/forum/topic/97645-date-to-unix-timestamp-conversion/
https://www.autoitscript.com/forum/topic/83667-epoch-time/

Which I should use in: `DemoCookies()` ?

Danp2:
> Is it sufficient to confirm how it works?

Yes, that matches my understanding from the official specs

wd_helper.au3
```diff
@@ -2180,6 +2180,57 @@ Func _WD_JsonActionPause($iDuration)
...
+; Parameters ....: $sName               - a string value. The name of the cookie.
+;                  $sValue              - a string value. The cookie value.
+;                  $sPath               - [optional] a string value. Default is '/'. This defines the cookie path.
+;                  $sDomain             - [optional] a string value. Default is '' - will be ommited and current browsing context’s of active document’s will be used. This defines the domain the cookie is visible to.
+;                  $bSecure             - [optional] a boolean value. Default is False. This defines whether the cookie is a secure cookie.
+;                  $bHTTPOnly           - [optional] a boolean value. Default is False. This defines whether the cookie is an HTTP only cookie.
+;                  $iExpiryTime         - [optional] an integer value. Default is 0 - will be ommited. This defines when the cookie expires, specified in seconds since Unix Epoch.
+;                  $sSameSite           - [optional] a string value. Default will be ommited and this mean "None". This defines whether the cookie applies to a SameSite policy. One of the following modes:
```
Danp2, line 2194: Please update to conform with rest of UDF. Should be like

```autoit
;                  $sValue              - The cookie value.
;                  $sPath               - [optional] This defines the cookie path. Default is '/'. 
```

wd_helper.au3
```diff
@@ -2180,6 +2180,57 @@ Func _WD_JsonActionPause($iDuration)
...
+
+	Local $vData = Json_ObjCreate()
+	Json_Put($vData, '.cookie.name', $sName)
+	Json_Put($vData, '.cookie.value', $sValue)
+	If $sPath Then Json_Put($vData, '.cookie.path', $sPath)
+	If $sDomain Then Json_Put($vData, '.cookie.domain', $sDomain)
+	If $bSecure Then Json_Put($vData, '.cookie.secure', $bSecure)
+	If $bHTTPOnly Then Json_Put($vData, '.cookie.httponly', $bHTTPOnly)
+	If $iExpiryTime Then Json_Put($vData, '.cookie.expiry', $iExpiryTime)
+	If $sSameSite Then Json_Put($vData, '.cookie.sameSite', $sSameSite)
```
Danp2, line 2223: I believe this would be better implemented if you did it like this --

-  Eliminate the first section of `If $variable = Default` checks
- Change the second section of If statements to `If $variable <> Default`

This will allow for doing things like passing False for $bSecure and having it correctly added to the cookie

mlipok: Take a look: [_WD_JsonCookie() inner validation](https://github.com/Danp2/au3WebDriver/pull/194/commits/6e4896a09377b54c2a51ed96ebef4576772545b5)

Will this be fine in this way ?

wd_helper.au3
mlipok: Take a look: [_WD_JsonCookie() parameters descriptions](https://github.com/Danp2/au3WebDriver/pull/194/commits/30710af72b92ef28f2d5710205a07c59bdb48bc9)

Danp2: $sName & $sValue still needs to be done.

wd_helper.au3
Danp2: The idea was that the optional parameters didn't generate any entries unless they were supplied. This change doesn't accomplish that.

wd_helper.au3
mlipok: Please check: [_WD_JsonCookie param descr fix](https://github.com/Danp2/au3WebDriver/pull/194/commits/6a24a523b393f1332088e6ea1db07afb657ebd76)

wd_helper.au3
mlipok: Default in the meaning of WebDriver spec is to omnit so you check `If $variable <> Default` and `Then Json_Put($vData, '.cookie.`
This way it works like intended by WebDriver spec .
But in this case (your modification) this is user resposibility to provide valid data.

And this is your intentionall choices of how this functions should works?

mlipok: > And this is your intentionall choices of how this functions should works?

If so, I can agree/aprove this way.

Danp2: > This way it works like intended by WebDriver spec

Unsure what you mean by this. 

IMO, this UDF isn't responsible for supplying default values to match the W3C specs. These entries are optional. When missing, it the webdriver's responsibility to supply any default values.

> But in this case (your modification) this is user resposibility to provide valid data.

As I said, additional checking for variable types could be added if desired. 

> And this is your intentionall choices of how this functions should works?

This is your function. I'm simply suggesting what I believe would be a better implementation where the `.cookie` entries will only be added for user-supplied values.

mlipok: > > This way it works like intended by WebDriver spec
> 
> Unsure what you mean by this.
> 
> IMO, this UDF isn't responsible for supplying default values to match the W3C specs. These entries are optional. When missing, it the webdriver's responsibility to supply any default values.

I mean that https://www.w3.org/TR/webdriver/#dfn-table-for-cookie-conversion says that if this parameters are omnited then WebDriver use the defaults by itself.

Danp2: Right.... those defaults get applied by the webdriver in the event the entry is omitted. Therefore, it isn't necessary for _WD_JsonCookie to handle these defaults.

Agreed?

mlipok: > This is your function. 

Oh indeed.
But you are the captain on this ship.

> I'm simply suggesting what I believe would be a better implementation where the `.cookie` entries will only be added for user-supplied values.

And here you use your captain's privileges.

Overall, it doesn't matter ....
It is important that I agree with your implementation / modifications.

wd_helper.au3
```diff
@@ -2180,6 +2180,52 @@ Func _WD_JsonActionPause($iDuration)
...
+;                  $sValue              - The cookie value.
+;                  $sPath               - [optional] This defines the cookie path. Default is '/'.
+;                  $sDomain             - [optional] This defines the domain the cookie is visible to. Default is '' - will be ommited and current browsing context’s of active document’s will be used.
+;                  $bSecure             - [optional] This defines whether the cookie is a secure cookie. Default is False.
+;                  $bHTTPOnly           - [optional] This defines whether the cookie is an HTTP only cookie. Default is False.
+;                  $iExpiryTime         - [optional] This defines when the cookie expires, specified in seconds since Unix Epoch. Default is 0 - will be ommited.
+;                  $sSameSite           - [optional] This defines whether the cookie applies to a SameSite policy. Default is '' - will be ommited and this mean "None". One of the following modes can be used:
+;                  |None
+;                  |Lax
+;                  |Strict
```
Danp2, line 2197: Please remove the "Default is" portions as the function no longer supplied default values. Same thing with the "will be omitted" portions.

Danp2 (approved):

Issue #197, opened 2022-02-14
_WD_UpdateDriver - Don't Delete WebDriver unless new Driver exists


seadoggie01:
Unfortunately, MS broke the normal upgrade path for WebDrivers today. The current "stable" version has only been made available for Mac users(!?). Because of this, _WD_UpdateDriver detected an update was available, attempted to download the new driver, failed, deleted the original driver, and returned an error.

Currently, this functionality can be seen by using `_WD_UpdateDriver("msedge", @AppDataDir & "\WebDrivers\", True, True)`

Edit: I'm looking into a way to fix this

seadoggie01:
The commit mlipok added fixes the issue, it returns False and sets @error = 13 (Not Found) instead of deleting the WebDriver and causing `$FilesInZip = $oShell.NameSpace($sTempFile).items` to cause an unhandle-able AutoIt due to a known issue in AutoIt.

mlipok:
> causing `$FilesInZip = $oShell.NameSpace($sTempFile).items` to cause an unhandle-able AutoIt due to a known issue in AutoIt.
Take a look here:
[$oNameSpace - fix for unhandled AutoIt error](https://github.com/Danp2/au3WebDriver/pull/198/commits/0a91968b54951055491efc5ab0ec0c8d9f8f11cb)

seadoggie01:
Yes, that would fix that error. The error only occurred because the downloaded file didn't exist, causing `$oShell.NameSpace($sTempFile)` to be null, then calling `.items` on it. So, theoretically, this shouldn't ever happen again, but probably better to prevent other unseen errors in the future :)

PR #198, opened 2022-02-14
_WD_UpdateDriver() check _WD_DownloadFile() result

mlipok:
Fix for: https://github.com/Danp2/au3WebDriver/issues/197

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,8 @@ Go to [legend](#legend---types-of-changes) for further information about the typ
 
 ### Fixed
 
+- _WD_UpdateDriver: Detect error during download and addtional error handling
+
 ### Removed
 
 ## [0.5.2] - 2022-02-11
--- a/wd_helper.au3
+++ b/wd_helper.au3
@@ -1383,39 +1383,42 @@ Func _WD_UpdateDriver($sBrowser, $sInstallDir = Default, $bFlag64 = Default, $bF
 				ElseIf $bUpdateAvail Or $bForce Then
 					$sTempFile = _TempFile($sInstallDir, "webdriver_", ".zip")
 					_WD_DownloadFile($sURLNewDriver, $sTempFile)
-
-					; Close any instances of webdriver and delete from disk
-					__WD_CloseDriver($sDriverEXE)
-					FileDelete($sInstallDir & $sDriverEXE)
-
-					; Handle COM Errors
-					Local $oErr = ObjEvent("AutoIt.Error", __WD_ErrHnd)
-					#forceref $oErr
-
-					; Extract new instance of webdriver
-					$oShell = ObjCreate("Shell.Application")
-					If @error Then
-						$iErr = $_WD_ERROR_GeneralError
+					If @error then
+						$iErr = @error
 					Else
-						$FilesInZip = $oShell.NameSpace($sTempFile).items
+						; Close any instances of webdriver and delete from disk
+						__WD_CloseDriver($sDriverEXE)
+						FileDelete($sInstallDir & $sDriverEXE)
+
+						; Handle COM Errors
+						Local $oErr = ObjEvent("AutoIt.Error", __WD_ErrHnd)
+						#forceref $oErr
+
+						; Extract new instance of webdriver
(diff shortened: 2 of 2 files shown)


Danp2:
@mlipok I'm going to assume that your fix is good since both you and @seadoggie01 tested it. Only issue I can see is that _WD_DownloadFile returns an error code of $_WD_ERROR_NotFound, which is not listed in the header for _WD_UpdateDriver. Either the header needs to be updated or the error code should be set via something like `$iErr = $_WD_ERROR_GeneralError` instead of `$iErr = @error`

mlipok:
Is that all now ?

Danp2:
AFAIK. Out of curiosity, I see `$_WD_ERROR_NoMatch` in the header. Do you happen to know under what conditions this would be triggered?

mlipok:
> Do you happen to know under what conditions this would be triggered?

interesting.... there was a bug in 

```AutoIt
Func _WD_DownloadFile($sURL, $sDest, $iOptions = Default)
	..........
	Local $sData = InetRead($sURL, $iOptions)
	If @error Then $iErr = $_WD_ERROR_GeneralError

		Local $hFile = FileOpen($sDest, $FO_OVERWRITE + $FO_BINARY)

		If $hFile <> -1 Then
			FileWrite($hFile, $sData)
			FileClose($hFile)

			$bResult = True
		Else
		EndIf
	Else
		$iErr = $_WD_ERROR_NotFound
	EndIf
	..........
EndFunc   ;==>_WD_DownloadFile
````

as in current `_WD_DownloadFile` implementation $_WD_ERROR_NotFound will never happen.

**Comment by mlipok** [comment]
I'm thinking about this kind of change:

```AutoIt
Func _WD_DownloadFile($sURL, $sDest, $iOptions = Default)
	..........
	Local $sData = InetRead($sURL, $iOptions)
	If @error Then $iErr = $_WD_ERROR_NotFound

		Local $hFile = FileOpen($sDest, $FO_OVERWRITE + $FO_BINARY)

		If $hFile <> -1 Then
			..........
		Else
		EndIf
	EndIf
	..........
EndFunc   ;==>_WD_DownloadFile
```

Danp2:
> as in current `_WD_DownloadFile` implementation, the `$_WD_ERROR_NotFound` error will never happen.

Actually, doesn't it first assign $_WD_ERROR_GeneralError and then subsequently assign $_WD_ERROR_NotFound when the If evaluates to False?

AFAICS, your proposed change should be fine. Still curious about `$_WD_ERROR_NoMatch`. 😉

mlipok:
> Actually, doesn't it first assign $_WD_ERROR_GeneralError and then subsequently assign $_WD_ERROR_NotFound when the If evaluates to False?

Oh. I missed it. You're right.

> AFAICS, your proposed change should be fine. 

Take a look: [$_WD_ERROR_NotFound fix](https://github.com/Danp2/au3WebDriver/pull/198/commits/ab19cdfd0168ad921d6d2c23f899231bf2f86ab6)

> Still curious about `$_WD_ERROR_NoMatch`. 😉

?

Danp2:
> Out of curiosity, I see $_WD_ERROR_NoMatch in the header. Do you happen to know under what conditions this would be triggered?

^ This is what I originally wrote above, but you didn't address it. To be clear, I was referring to the header for _WD_UpdateDriver

mlipok:
This could also happen as a result of `_WD_GetBrowserVersion` or `_WD_GetWebDriverVersion`

Danp2:
> This could also happen as a result of _WD_GetBrowserVersion or _WD_GetWebDriverVersion

Those functions return `$_WD_ERROR_NotFound`, which is different from `$_WD_ERROR_NoMatch`

Am I missing something?!

mlipok:
Yeah.
That I am tired today and needed a little more time to understand what you were asking for.

mlipok:
> Out of curiosity, I see `$_WD_ERROR_NoMatch` in the header. Do you happen to know under what conditions this would be triggered?

Was trigered last time here: https://github.com/Danp2/au3WebDriver/commit/a3b1a894eb7238d1e127aba4224dccd57760e1ba#r66697914

PR #200, opened 2022-02-15
added: _WD_GetDevicePixelRatio()

mlipok:


Sven-Seyfert (approved):
Looks good to me, that's why I approve the PR.

**One question left for me:**
I see no real purpose of having this ratio value. What will the user do with this kind of information?

> I though it's only interesting for developers who write web applications. They possibly want to have different site behavior in case of different pixel ratio (different image rendering or the usage of a different font, for example).

What is the advantage for a user who **automates** a browser actions?

mlipok:
> What is the advantage for a user who **automates** a browser actions?

https://www.autoitscript.com/forum/topic/205553-webdriver-udf-help-support-iii/?do=findComment&comment=1493911

wd_helper.au3
```diff
@@ -1998,6 +1998,28 @@ Func _WD_IsFullScreen($sSession)
...
+; Remarks .......:
+; Related .......:
+; Link ..........: https://developer.mozilla.org/en-US/docs/Web/API/Window/devicePixelRatio
+; Example .......: No
+; ===============================================================================================================================
+Func _WD_GetDevicePixelRatio($sSession)
+	Local Const $sFuncName = "_WD_GetDevicePixelRatio"
+	Local $sResponse = _WD_ExecuteScript($sSession, "return window.devicePixelRatio", Default, Default, $_WD_JSON_Value)
+	Local $iErr = @error
+	If $iErr Then $sResponse = 0
```
Danp2, line 2019: This wipes out the response from the WebDriver, which is different behavior than stated in the header. Which behavior is correct?

mlipok: Fixed

PR #234, opened 2022-03-01
OPERA: _WD_GetBrowserVersion refactoring

mlipok:
### Proposed changes

For me it is resonable/more rational that `If FileExists($sBrowser) Then` is checked first.
I know that this will be not comonly used case (file checking instead registry checking) but this function is used once per script run and there is nothing to do with performance, only with code clarity and error handling.
Thus I also added few `$_WD_ERROR_FileIssue` and `$iExt` and also I propose to add `$_WD_ERROR_NotSupported`

### Checklist

### Types of changes

- [ ] Other (please describe)

Not optimal code (in my opinin).

### What is the new behavior?

Better code clarity , and better error handling

### Additional context

NONE

### System under test

NOT RELATED

### DIFF
--- a/wd_helper.au3
+++ b/wd_helper.au3
@@ -1429,8 +1429,11 @@ Func _WD_GetBrowserVersion($sBrowser)
 	Local $iErr = $_WD_ERROR_Success, $iExt = 0
 	Local $sBrowserVersion = "0"
 
-	If FileExists($sBrowser) Then
-		If _WinAPI_GetBinaryType($sBrowser) = 0 Then
+	If StringInStr($sBrowser,'\') Then
+		If FileExists($sBrowser) = 0 Then
+			$iErr = $_WD_ERROR_FileIssue
+			$iExt = 21
+		ElseIf _WinAPI_GetBinaryType($sBrowser) = 0 Then
 			$iErr = $_WD_ERROR_FileIssue
 			$iExt = 22
 		Else
--- a/wd_helper.au3
+++ b/wd_helper.au3
@@ -1426,32 +1426,41 @@ EndFunc   ;==>_WD_UpdateDriver
 ; ===============================================================================================================================
 Func _WD_GetBrowserVersion($sBrowser)
 	Local Const $sFuncName = "_WD_GetBrowserVersion"
-	Local $iErr = $_WD_ERROR_Success
+	Local $iErr = $_WD_ERROR_Success, $iExt = 0
 	Local $sBrowserVersion = "0"
 
-	Local $sPath = _WD_GetBrowserPath($sBrowser)
-
-	If @error Then
-		$iErr = $_WD_ERROR_NotFound
-
-		If FileExists($sBrowser) Then
-			; Directly retrieve file version if full path was supplied
-			$sBrowserVersion = FileGetVersion($sBrowser)
-
-			If Not @error Then
-				; Extract filename and confirm match in list of supported browsers
-				$sBrowser = StringRegExpReplace($sBrowser, "^.*\\|\..*$", "")
-				If _ArraySearch($_WD_SupportedBrowsers, $sBrowser, Default, Default, Default, Default, Default, $_WD_BROWSER_Name) <> -1 Then _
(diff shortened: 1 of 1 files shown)


Danp2:
> You reversed the order of checking FileExists and _WD_GetBrowserPath. Doing it this way can lead to problems. For example, what happens if the working directory contains a file named chrome?

I previously tried to explain why this wasn't a good way to do it. Unsure if you didn't read my comment or just didn't understand it. So let us try this again. Run the following code with your revision --

```autoit
$sFile = @ScriptDir & "\chrome"
FileDelete($sFile)
$sVersion = _WD_GetBrowserVersion('chrome')
ConsoleWrite('@@ Debug(' & @ScriptLineNumber & ') : $sVersion = ' & $sVersion & @CRLF & '>Error code: ' & @error & @CRLF) ;### Debug Console
FileWrite($sFile, "test")
$sVersion = _WD_GetBrowserVersion('chrome')
ConsoleWrite('@@ Debug(' & @ScriptLineNumber & ') : $sVersion = ' & $sVersion & @CRLF & '>Error code: ' & @error & @CRLF) ;### Debug Console
```
What results did you observe?

mlipok:
```autoit
_WD_GetBrowserPath ==> Success
_WD_GetBrowserVersion ==> Success
@@ Debug(66) : $sVersion = 98.0.4758.102
>Error code: 0
_WD_GetBrowserVersion ==> File issue
@@ Debug(69) : $sVersion = 0.0.0.0
>Error code: 18
````

**Comment by mlipok** [comment]
after [_WinAPI_GetBinaryType checking](https://github.com/Danp2/au3WebDriver/pull/234/commits/1375ae56b5f9d7859144d2d77bdf2f7284b5a228)

```AutoIt
_WD_GetBrowserPath ==> Success
_WD_GetBrowserVersion ==> Success
@@ Debug(66) : $sVersion = 98.0.4758.102
>Error code: 0
_WD_GetBrowserVersion ==> File issue
@@ Debug(69) : $sVersion = 0
>Error code: 18
```

Danp2:
Now run the same code snippet against the `supportOpera` code base and post the results.

Danp2:
I believe that you are still missing the point that your version breaks when a file exists that matches the browser name. The intent was to pass the browser name to the UDF, `chrome` in this case. The UDF shouldn't fail to return the corrrect value just because an arbitrary file with the same name exists in the current working directory.

> And it's not a good thing that the UDF returns based on the registry instead of on the basis of the file........ I mean in case when user/coder pass file then UDF should return values based on the file not on the registry.

Agreed. Can you show an example where the current implementation in `supportOpera` doesn't follow this logic?

mlipok:
> Can you show an example where the current implementation in `supportOpera` doesn't follow this logic?

I was not able because it wors fine.

After my PR [check if param contains "\"](https://github.com/Danp2/au3WebDriver/pull/234/commits/b522836de3e1452e43d7e00f2e377dc7de96cee5)

This also fulfills your condition:
> The UDF shouldn't fail to return the corrrect value just because an arbitrary file with the same name exists in the current working directory.

but at the same time have better error handling.

mlipok:
And what about `$_WD_ERROR_NotSupported` ?

Danp2:
> I was not able because it works fine.

Glad to hear it. 

> Can you agree with my proposal in the current state ?

No, because it isn't a better implementation IMO. You appear to be fixated on the order of events. I'm more concerned about actual functionality. 

I can still envision scenarios where your code would fail. For example --

```autoit
	FileChangeDir("C:\Program Files\Google\Chrome\Application")
	$sVersion = _WD_GetBrowserVersion("chrome.exe")
	ConsoleWrite('@@ Debug(' & @ScriptLineNumber & ') : $sVersion = ' & $sVersion & @CRLF & '>Error code: ' & @error & @CRLF) ;### Debug Console
```

While it might be nice to perform `FileExist` checks first, it presents too many issues that don't exist with the current implementation.

mlipok:
I see your point now. So I will propose here implementation with yours `order of events` but with my error handling then we will talk again about the code.

Are you ok with that ?

PR #235, opened 2022-03-01
OPERA:  _WD_GetBrowserVersion/Path refactoring

mlipok:
### Proposed changes

I have added more error handling and fixed `_WD_UpdateDriver` in case when file full path is used `Extract filename from full path`

```AutoIt
			If StringInStr($sBrowser, '\') Then
				; Extract filename from full path
				$sBrowser = StringRegExpReplace($sBrowser, "^.*\\|\..*$", "")
			EndIf

			; Match in list of supported browsers
			Local $iIndex = _ArraySearch($_WD_SupportedBrowsers, $sBrowser, Default, Default, Default, Default, Default, $_WD_BROWSER_Name)
			$sDriverEXE = $_WD_SupportedBrowsers[$iIndex][$_WD_BROWSER_DriverName]
```

### Checklist

### Types of changes

- [ ] Other (please describe)

Error handling is not complete.
`_WD_UpdateDriver` may fail in case when file full path is used

### What is the new behavior?

better error handling and fixed `_WD_UpdateDriver` in case when file full path is used

### Additional context

https://github.com/Danp2/au3WebDriver/pull/234

### System under test

NOT RELATED

### DIFF
--- a/wd_helper.au3
+++ b/wd_helper.au3
@@ -1302,15 +1302,15 @@ Func _WD_UpdateDriver($sBrowser, $sInstallDir = Default, $bFlag64 = Default, $bF
 
 		$sBrowserVersion = _WD_GetBrowserVersion($sBrowser)
 		$iErr = @error
+		Local $iIndex = @extended
 
 		If $iErr = $_WD_ERROR_Success Then
-			If StringInStr($sBrowser, '\') Then 
-				; Extract filename from full path
+			If StringInStr($sBrowser, '\') Then
+				; Extract filename from full path - for any case
 				$sBrowser = StringRegExpReplace($sBrowser, "^.*\\|\..*$", "")
 			EndIf
 
-			; Match in list of supported browsers
-			Local $iIndex = _ArraySearch($_WD_SupportedBrowsers, $sBrowser, Default, Default, Default, Default, Default, $_WD_BROWSER_Name)
+			; Match exe file name in list of supported browsers
 			$sDriverEXE = $_WD_SupportedBrowsers[$iIndex][$_WD_BROWSER_DriverName]
 
 			; Determine current local webdriver Architecture
--- a/wd_helper.au3
+++ b/wd_helper.au3
@@ -1438,10 +1438,9 @@ Func _WD_GetBrowserVersion($sBrowser)
 	Local $sBrowserVersion = "0"
 
 	Local $sPath = _WD_GetBrowserPath($sBrowser)
+	$iErr = @error
+	$iExt = @extended
 	If @error Then
-		$iErr = @error
-		$iExt = @extended
-
 		; as registry checks fails, now checking if file exist and is exeutable
 		If FileExists($sBrowser) Then
 			If _WinAPI_GetBinaryType($sBrowser) = 0 Then
@@ -1454,9 +1453,11 @@ Func _WD_GetBrowserVersion($sBrowser)
 
 				; Extract filename and confirm match in list of supported browsers
 				$sBrowser = StringRegExpReplace($sBrowser, "^.*\\|\..*$", "")
-				If _ArraySearch($_WD_SupportedBrowsers, $sBrowser, Default, Default, Default, Default, Default, $_WD_BROWSER_Name) = -1 Then
+				Local $iIndex = _ArraySearch($_WD_SupportedBrowsers, $sBrowser, Default, Default, Default, Default, Default, $_WD_BROWSER_Name)
+				If $iIndex = -1 Then
 					$iErr = $_WD_ERROR_NotSupported
 				Else
+					$iExt = $iIndex
 					$sPath = $sBrowser
 				EndIf
 			EndIf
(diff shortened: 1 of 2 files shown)


wd_helper.au3
```diff
@@ -1489,12 +1512,21 @@ Func _WD_GetBrowserPath($sBrowser)
...
 		Else
 			$sPath = StringRegExpReplace($sPath, '["'']', '') ; Remove quotation marks
 			$sPath = StringRegExpReplace($sPath, '(.+\\)(.*exe)', '$1' & $sEXE) ; Registry entries can contain "Launcher.exe" instead "opera.exe"
+
+			If FileExists($sPath) = 0 Then
+				$iErr = $_WD_ERROR_FileIssue
+				$iExt = 23
+			ElseIf _WinAPI_GetBinaryType($sPath) = 0 Then
+				$iErr = $_WD_ERROR_FileIssue
+				$iExt = 24
```
Danp2, line 1525: I'm not sure how I feel about this. This intent of this function was to `Retrieve path to browser executable from registry` (taken directly from the header description). It doesn't state anything about validating that the resulting file path exists on disk.

Please explain why you think this code belongs here.

wd_helper.au3
```diff
@@ -1303,6 +1304,12 @@ Func _WD_UpdateDriver($sBrowser, $sInstallDir = Default, $bFlag64 = Default, $bF
 		$iErr = @error
 
 		If $iErr = $_WD_ERROR_Success Then
+			If StringInStr($sBrowser, '\') Then 
+				; Extract filename from full path
+				$sBrowser = StringRegExpReplace($sBrowser, "^.*\\|\..*$", "")
+			EndIf
+
+			; Match in list of supported browsers
 			Local $iIndex = _ArraySearch($_WD_SupportedBrowsers, $sBrowser, Default, Default, Default, Default, Default, $_WD_BROWSER_Name)
```
Danp2, line 1313: What do you think about having _WD_GetBrowserVersion set @extended to the matching index (obviously only when the function is successful)? That would eliminate the need to perform this extraction again or even perform the _ArraySearch again.

wd_helper.au3
mlipok: Function name is `_WD_GetBrowserPath` 
When a function gives a path to a nonexistent file, it should mark it up somehow. The same if the file is not correct (for example not executable).
Everyone expects the function to return a valid working file (path to such a file).

btw. 
[You give me the reason](https://github.com/Danp2/au3WebDriver/pull/234#issuecomment-1055491099) to add such feature.
`The UDF shouldn't fail to return the corrrect value just because an arbitrary file with the same name exists in the current working directory.`

I hope you find my reasoning valuable.

wd_helper.au3
mlipok: Good point.
I'll make the appropriate changes

wd_helper.au3
Danp2: You may expect it, but I doubt that everyone does. 🙂 

Also, there is a difference between failing to return the correct value and performing checks beyond the scope of the function.

mlipok: If we are adding support for a portable version of browsers, we have to predict that the user will enter the wrong path. Wrong in any way. The UDF should report it so that you can easily find out the cause of the error. In this respect, it is not beyond the scope of this function.

wd_helper.au3
Danp2: Looks good to me

wd_helper.au3
Danp2: `_WD_GetBrowserPath` has nothing to do with supporting portable browsers, so unsure why you brought that up in the discussion. `_WD_GetBrowserVersion` is where these checks should be performed IMO, and the functionality is already in place.

mlipok: So take a look here: [WinAPI_GetBinaryType usage](https://github.com/Danp2/au3WebDriver/pull/235/commits/f2e41332e20c6674fae0520df863958624f0732f)

wd_helper.au3
```diff
@@ -1301,9 +1302,15 @@ Func _WD_UpdateDriver($sBrowser, $sInstallDir = Default, $bFlag64 = Default, $bF
...
 		$sBrowserVersion = _WD_GetBrowserVersion($sBrowser)
 		$iErr = @error
+		Local $iIndex = @extended
 
 		If $iErr = $_WD_ERROR_Success Then
-			Local $iIndex = _ArraySearch($_WD_SupportedBrowsers, $sBrowser, Default, Default, Default, Default, Default, $_WD_BROWSER_Name)
+			If StringInStr($sBrowser, '\') Then
+				; Extract filename from full path - for any case
+				$sBrowser = StringRegExpReplace($sBrowser, "^.*\\|\..*$", "")
+			EndIf
```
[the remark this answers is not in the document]
mlipok: Take a look on: [If StringInStr($sBrowser, '\') removal](https://github.com/Danp2/au3WebDriver/pull/235/commits/8c5303407150c68805dd96cf7d5921e106b6f64b)

Danp2 (approved):

Issue #219, opened 2022-02-24
mLipok plans for next months


mlipok:
### Is your feature request related to a problem? Please describe

Things to do after 'Opera' projects.
Please treat it as a notebook, sketch, draft, remainder.
Please do not analyze before finishing the 'Opera project'.
Please assign this ticket to me.
I just don't want to disrupt the ongoing design work for now, but I also don't want to forget what I want to do in the future.

### Describe the solution you'd like

- [x] 1. review: _WD_Option() for such kind of variables like $_WD_ERROR_MSGBOX which is not possible to set with _WD_Option()  >>> https://github.com/Danp2/au3WebDriver/pull/244 

- [x] 2. refactor: __WD_Error to use also $iExt - currently it use only $iErr >>> https://github.com/Danp2/au3WebDriver/pull/244

- [x] 3. magic number removall : `MsgBox(16, ` >>> https://github.com/Danp2/au3WebDriver/pull/244

- [X] 4a. __WD_ConsoleWrite : preserve `@error` and `@extended` >>> https://github.com/Danp2/au3WebDriver/pull/222

- [X] 4n. __WD_ConsoleWrite: move `@CRLF` to internall processing - not needed in all `__WD_ConsoleWrite` line >>> https://github.com/Danp2/au3WebDriver/pull/242

- [X]  4c. __WD_ConsoleWrite: support for other `@CRLF`  will be needed in case you have your own function defined .... `If IsFunc($_WD_CONSOLE) Then` >>> https://github.com/Danp2/au3WebDriver/pull/242

Example:
```autoit
Global $_WD_CONSOLE_SUFFIX = @crlf

Func __WD_ConsoleWrite($sMsg)
	If IsFunc($_WD_CONSOLE) Then
		Call($_WD_CONSOLE, $sMsg & $_WD_CONSOLE_SUFFIX)
	ElseIf $_WD_CONSOLE = Null Then
		; do nothing
	Else
		FileWrite($_WD_CONSOLE, $sMsg& $_WD_CONSOLE_SUFFIX)
	EndIf
EndFunc   ;==>__WD_ConsoleWrite
```

- [x]  5.  >>> https://github.com/Danp2/au3WebDriver/pull/244
`DllCall("kernel32.dll", "none", "OutputDebugString", "str", $sMsg)` can not be obligatory
It need to have an option to switch it off.
More: it should be switched off by design (get a look for GDPR law act).

- [x]  6.
[...]

PR #244, opened 2022-03-04
__WD_Error - improved error handling

mlipok:
### Proposed changes

This PR adds many feature to `__WD_Error()`

+ `$_WD_ERROR_UserAbort` added feature for user interaction `abort script runs` in `__WD_Error()`
+ `$_WD_ERROR_OUTPUTDEBUG` logs created by `DllCall("kernel32.dll", "none", "OutputDebugString", "str", $sMsg)` are disabled by default - could be Enabled/Disabled by `_WD_Option('OUTPUTDEBUG', False)`
+ `$_WD_ERROR_MSGBOX` - could be Enabed/Disabled by `_WD_Option('errormsgbox', False)`
+ `$iExt` - `__WD_Error()` is able to use $iExt
+ MagicNumber replaced with ENUMS in `"WD_Core.au3 Error:", $sMsg`

### Checklist

### Types of changes

- [x] Other (`OutputDebugString` is disabled by default. **So is it ScriptBreakingChange ?** or only requires Warrning in relase note ?)

lack of Ext inf and non configurable way of procesing in `__WD_Error()`

### What is the new behavior?

Configurable way of procesing in `__WD_Error()` with more information

### Additional context

(https://github.com/Danp2/au3WebDriver/issues/219)

### System under test

NOT RELATED


wd_core.au3
```diff
@@ -172,6 +173,7 @@ Global $_WD_DRIVER_CLOSE = True ; Close prior driver instances before launching
 Global $_WD_DRIVER_DETECT = True ; Don't launch new driver instance if one already exists
 Global $_WD_RESPONSE_TRIM = 100 ; Trim response string to given value for debug output
 Global $_WD_ERROR_MSGBOX = True ; Shows in compiled scripts error messages in msgboxes
+Global $_WD_ERROR_OUTPUTDEBUG = False ; Log errors to "OutputDebugString"
```
mlipok, line 176: `OutputDebugString` is disabled by default. So is it `ScriptBreakingChange` ? or only requires Warrning in relase note ?

Danp2: No, I wouldn't consider that script breaking.

wd_helper.au3
```diff
@@ -1061,7 +1061,7 @@ Func _WD_ElementSelectAction($sSession, $sSelectElement, $sCommand)
 		__WD_ConsoleWrite($sFuncName & ': ' & ((IsArray($vResult)) ? "(array)" : $vResult))
 	EndIf
 
-	Return SetError(__WD_Error($sFuncName, $iErr), $_WD_HTTPRESULT, $vResult)
+	Return SetError(__WD_Error($sFuncName, $iErr, Default, $iExt), $_WD_HTTPRESULT, $vResult)
```
Danp2, line 1064: undeclared variable

wd_core.au3
```diff
@@ -1588,15 +1608,30 @@ Func __WD_Error($sWhere, $i_WD_ERROR, $sMessage = Default)
...
+						EndIf
+					EndIf
+					If $_WD_ERROR_OUTPUTDEBUG Then
+						DllCall("kernel32.dll", "none", "OutputDebugString", "str", $sMsg)
+					EndIf
+				EndIf
+			EndIf
+
+			If $_WD_DEBUG = $_WD_DEBUG_Info And $iErr = $_WD_ERROR_UserAbort Then
+				__WD_ConsoleWrite($sFuncName & ': UserAbort on: ' & $sMsg)
```
Danp2, line 1630: This should go inside the prior If..Endif unless you are wanting this to handle when the incoming error is $_WD_ERROR_UserAbort

wd_core.au3
```diff
@@ -1588,15 +1608,30 @@ Func __WD_Error($sWhere, $i_WD_ERROR, $sMessage = Default)
 				$sMsg = $sMsg & ": " & $sMessage
 			EndIf
 
+			$sMsg &= (($iExt) ? (' : Extended = ' & $iExt) : (''))
```
Danp2, line 1611: I'm thinking this would be better if we showed the numeric value for both @error and @extended. Something like this (untested) --

`$sMsg &= '(' & $i_WD_ERROR & (($iExt) ? ' / ' &  $iExt) : '') & ')`

Danp2 (changes requested):

wd_helper.au3
mlipok: Fixed in: [_WD_ElementSelectAction $iExt fix](https://github.com/Danp2/au3WebDriver/pull/244/commits/1cacedb02289a53fb30b34956dec36b16d0963fb)
This was because I'm using version which already have `_WD_ElementSelectAction - multiple selection support PR#212`

wd_core.au3
mlipok: please take a look [moved: If $_WD_DEBUG = $_WD_DEBUG_Info Then](https://github.com/Danp2/au3WebDriver/pull/244/commits/b312dedc8b39a4e79e20050aaae7fa1315d7f7fa)

wd_core.au3
mlipok: Take a look: [EXT moved, added ERR numeric value](https://github.com/Danp2/au3WebDriver/pull/244/commits/def7c67a88f63ab16105ebe9638bd5ebaa3cd855)
What you think about such concept ?

Danp2: I would prefer something less wordy / intrusive. That's why I made the prior suggestion where the error codes were displayed in parentheses without further labeling. I even think we should take out the conditional display of @extended and always show the info like this --

`_WD_GetBrowserVersion ==> General Error (1 / 0)`

This improves the output of __WD_Error by always showing the error code, but it doesn't become too obtrusive.

mlipok: Take a look: [Func __WD_Error() - Refactoring](https://github.com/Danp2/au3WebDriver/pull/244/commits/4be951248cdadb096191db4f09d66427af7d0669)

Danp2: - What's point of adding $iErr and changing $i_WD_ERROR > $iErr? This doesn't appear to serve any real purpose.

- IIRC, the @compiled logic came over from FF.au3. I'm ok with removing this, but I wonder if it will cause a behavior change for existing scripts. Maybe `$_WD_ERROR_MSGBOX` should default to False?

- I would suggest further changing `WD_Core UDF Error` to `Webdriver UDF Error`

mlipok: > * What's point of adding $iErr and changing $i_WD_ERROR > $iErr? This doesn't appear to serve any real purpose.

I can revert it, if you wanted?

> * IIRC, the @compiled logic came over from FF.au3. I'm ok with removing this, but I wonder if it will cause a behavior change for existing scripts. Maybe `$_WD_ERROR_MSGBOX` should default to False?

Good idea about:
```AutoIt
Global $_WD_ERROR_MSGBOX = False ; Shows in compiled scripts error messages in msgboxes
```

Yes it change behavior. On compiled scirpt there was user interaction required when `MsgBox()` popups.
Now it will not be required by default. Unless sombody enable them.

I think that after all we need to suplement `wd_demo.au3` by adding this:
```autoit
_WD_Option("errormsgbox", (https://github.com/compiled =1))
_WD_Option("OutputDebug", (https://github.com/compiled =1))
```

> * I would suggest further changing `WD_Core UDF Error` to `Webdriver UDF Error`

Good point.

mlipok: as to `$iErr` and `$i_WD_ERROR`

```autoit
Func __WD_Error($sWhere, $i_WD_ERROR, $sMessage = Default, $iExt = 0)
	.......
	Local $sMsg, $iErr = $i_WD_ERROR ; as we set here $iErr = $i_WD_ERROR
	.......
	Switch $_WD_DEBUG
		.......
		Case $_WD_DEBUG_Error
			.......
		Case $_WD_DEBUG_Info
			.......
				If $_WD_ERROR_MSGBOX And $iErr < 6 Then
					.......
					If $iAnswer = $IDCANCEL Then
						$iErr = $_WD_ERROR_UserAbort ; evenutally change here
						.......
					EndIf
				EndIf
			.......
	EndSwitch

	Return $iErr ; and return here
EndFunc   ;==>__WD_Error
```

I was thinking that this is no more needed to use `$i_WD_ERROR`

After rethinking we can also use this function to not use `$i_WD_ERROR` variable at all and stick with commonly used name `$iErr`:

```autoit
Func __WD_Error($sWhere, $iErr, $sMessage = Default, $iExt = 0)
	.......
	Local $sMsg
	.......
	Switch $_WD_DEBUG
		.......
		Case $_WD_DEBUG_Error
			.......
		Case $_WD_DEBUG_Info
			.......
				If $_WD_ERROR_MSGBOX And $iErr < 6 Then
					.......
					If $iAnswer = $IDCANCEL Then
						$iErr = $_WD_ERROR_UserAbort ; evenutally change here
						.......
					EndIf
				EndIf
			.......
	EndSwitch

	Return $iErr ; and return here
EndFunc   ;==>__WD_Error
```

Danp2: > After rethinking we can also use this function to not use $i_WD_ERROR variable at all and stick with commonly used name $iErr

Yes, that's something that I had considered as well. IMO, this change isn't needed but it shouldn't hurt to change this.

mlipok: Take a look: [further refactoring](https://github.com/Danp2/au3WebDriver/pull/244/commits/046de6edf6a3404fdfb67e077d64aca85148d273)

Danp2: > `$iExt = 0`

I suggest that this be changed to `$iExt = Default` in the function declaration. Then you can distinguish between @extended truly equal to 0 versus not being passed from the calling routine.

wd_core.au3
```diff
@@ -1589,11 +1589,12 @@ EndFunc   ;==>__WD_Delete
 ; Link ..........:
 ; Example .......: No
 ; ===============================================================================================================================
-Func __WD_Error($sWhere, $iErr, $sMessage = Default, $iExt = 0)
+Func __WD_Error($sWhere, $iErr, $sMessage = Default, $iExt = Default)
 	Local Const $sFuncName = "__WD_Error"
 	Local $sMsg
 
 	If $sMessage = Default Then $sMessage = ''
+	If $iExt = Default Then $iExt = 0
```
mlipok, line 1597: currently this is needed, maybe in future we will check other cases

wd_core.au3
mlipok: Take a look again on: [again refactoring](https://github.com/Danp2/au3WebDriver/pull/244/commits/ab2a021c3cdc14236114679ba73ea69408134b54)

wd_core.au3
Danp2: I suggest that you leave it set to `default`. Then you could change this line --

`Local $sExtended = (($iExt) ? (" / " & $iExt) : (""))`

to

`Local $sExtended = (($iExt <> Default) ? (" / " & $iExt) : (""))`

wd_core.au3
```diff
@@ -1557,46 +1575,59 @@ EndFunc   ;==>__WD_Delete
...
+			Local $sExtended = (($iExt) ? (" / " & $iExt) : (""))
+			$sMsg = $sWhere & " ==> " & $aWD_ERROR_DESC[$iErr] & " ( " & $iErr & $sExtended & " )"
+			$sMsg &= (($sMessage) ? (" : " & $sMessage) : (""))
 			__WD_ConsoleWrite($sMsg)
 
-			If @Compiled Then
-				If $_WD_ERROR_MSGBOX And $i_WD_ERROR <> $_WD_ERROR_Success And $i_WD_ERROR < 6 Then MsgBox(16, "WD_Core.au3 Error:", $sMsg)
-				DllCall("kernel32.dll", "none", "OutputDebugString", "str", $sMsg)
+			If $iErr <> $_WD_ERROR_Success Then
+				If $_WD_ERROR_MSGBOX And $iErr < 6 Then
```
Danp2, line 1612: I think we need to review this. Instead of "6", we could use the defined constant. Also, does this logic still apply? If so, then is 6 still the correct cutoff or has it changed as the UDF has grown?

wd_core.au3
```diff
@@ -1557,46 +1575,59 @@ EndFunc   ;==>__WD_Delete
...
+					DllCall("kernel32.dll", "none", "OutputDebugString", "str", $sMsg)
+				EndIf
 			EndIf
+
 	EndSwitch
 
-	Return $i_WD_ERROR
+	; $iErr should be returned as "Return Value" as it is used in the calling function like this:     Return SetError(__WD_Error(...))
+	; $iExt should be returned as @extended as it can be used in the calling function like this:         Return SetError(__WD_Error(..., $iExt), @extended, ....)
+	Return SetError($iErr, $iExt, $iErr)
```
Danp2, line 1630: Not sure about this, especially since you've already acknowledged in the comment two lines above about using this function as a parameter to SetError.

wd_core.au3
mlipok: I go to bed.
I will answer tomorrow.

mlipok: I agree that this is things to review but not in this PR
So I create:
https://github.com/Danp2/au3WebDriver/issues/246

Could you take care of this problem this time?

wd_core.au3
mlipok: Take a look on: [comments on the end](https://github.com/Danp2/au3WebDriver/pull/244/commits/88e47dfdbcf4dda93d197ad5326a67d65ecb2c92)

wd_core.au3
mlipok: Take a look on: [$iExt Default refactoring](https://github.com/Danp2/au3WebDriver/pull/244/commits/244d9ee0b0292fed21cf5c22aade1491c1219433)

wd_core.au3
```diff
@@ -1603,14 +1602,14 @@ Func __WD_Error($sWhere, $iErr, $sMessage = Default, $iExt = Default)
...
-			Local $sExtended = (($iExt) ? (" / " & $iExt) : (""))
+			Local $sExtended = (($iExt <> Default) ? (" / " & $iExt) : (""))
 			$sMsg = $sWhere & " ==> " & $aWD_ERROR_DESC[$iErr] & " ( " & $iErr & $sExtended & " )"
 			$sMsg &= (($sMessage) ? (" : " & $sMessage) : (""))
 			__WD_ConsoleWrite($sMsg)
 
 			If $iErr <> $_WD_ERROR_Success Then
 				If $_WD_ERROR_MSGBOX And $iErr < 6 Then
-					Local $iAnswer = MsgBox($MB_ICONERROR + $MB_OKCANCEL, "Webdriver UDF Error:", $sMsg)
+					Local $iAnswer = MsgBox($MB_ICONERROR + $MB_OKCANCEL, "au3WebDriver UDF Error:", $sMsg)
```
Danp2, line 1612: au3WebDriver is the repo name. I still use WebDriver when referring to the UDF. 😉

mlipok: Take a look: [au3WebDriver > WebDriver](https://github.com/Danp2/au3WebDriver/pull/244/commits/5c3cb79b7b1a44e398daf36834502554d16813fc)

wd_core.au3
Danp2: My earlier comment was in reference to changing the line `	Return $i_WD_ERROR` to `Return SetError($iErr, $iExt, $iErr)`. So the issue is with adding SetError, not with your comments.

It doesn't make sense to me to add the call to SetError here when we know that it is expected that __WD_Error is being passed as a parameter to SetError at he end of each _WD* function.

mlipok: Take a look on: [Return $iErr reverted](https://github.com/Danp2/au3WebDriver/pull/244/commits/db5a376bcab3b9cb0bc8de41c149a14aa544deaf)

Danp2 (approved):

PR #274, opened 2022-03-29
GDPR (law act) better accordance

mlipok:
### Proposed changes

Modify UDF so that the default settings are GDPR compliant when running compiled scripts.

### Checklist

### Types of changes

- [x] Other (please describe)

`$_WD_DEBUG = $_WD_DEBUG_Info`

### What is the new behavior?

Defalult value for `$_WD_DEBUG` vary on compilation status.

### Additional context

Any UDF shouldn't submit any message to console by default.
I mean when compiled program uses UDF.
It is user/programist/script responsibility to manage which data should be output and where it should be output.

### System under test

not related


mlipok:
In GDPR we have something which is called like "Security by default"
I mean that by default UDF should not put any messages to console when program is compiled.
When program is run from script without compilation then by default console messages should be disabled.


> 1. It could break an existing script where the developer is depending on the existing default value

I do not agree here.
Script will work the same way as currently.
Only loging is changed, but it does not affect the overall script execution.

It can be mention as a WARNING but it is not script breaking change.

Danp2:
> In GDPR law act we have (in Europe) something which is called like "Security by default"
I mean that by default UDF should not put any messages to console when program is compiled.

This type of details should have been included in your original PR as justification for the change. Please be sure to include these details in the future. Otherwise, I can only guess at why the proposal was made.

>  When program is run from script without compilation then by default console messages should be disabled.

Your proposed change wouldn't address this, right?

> I do not agree here. Script will work the same way as currently.

While the webdriver functions would continue to work correctly, there could be other scenarios where this change could affect an existing script. For example: The script could be checking for the console to appear and exit prematurely if not found.

I admit that the likelihood of this situation occurring is small, but it's still a possibility. 🙂

mlipok:
> > When program is run from script without compilation then by default console messages should be disabled.
> 
> Your proposed change wouldn't address this, right?

My mistake. I meant the opposite.

`When program is run from script without compilation then by default console messages can be enabled.` but when script is compiled then console messages should be disabled by default.

mlipok:
> I admit that the likelihood of this situation occurring is small, but it's still a possibility.

Agree. So should a WARNING in documentation be enough?

Danp2:
I suggest that we take a step back and determine which actions would be considered unacceptable under GDPR. Here's a list to start things off --

- Logging to a console?
- Logging to a file?
- Displaying the webdriver console?

Are there any others that would be of concern?

P.S. This whole thing would have been better off as an issue where we resolved these questions before it became a PR. 😉

mlipok:
> I suggest that we take a step back and determine

Ok.

Danp2:
> Here's a list to start things off --
> 
>     * Logging to a console?
> 
>     * Logging to a file?
> 
>     * Displaying the webdriver console?
> 
> 
> Are there any others that would be of concern?

@mlipok Were you planning to respond to this?

mlipok:
Yeah. Just have too many things to do.

Danp2:
This line should be reviewed as part of this PR. Also need to fix GDRP > GDPR

https://github.com/Danp2/au3WebDriver/blob/589262348ef3922230ff5938279bd717ca57a113/wd_capabilities.au3#L445

wd_capabilities.au3
```diff
@@ -423,7 +423,6 @@ EndFunc   ;==>__WD_CapabilitiesNotation
 ; Example .......: No
 ; ===============================================================================================================================
 Func _WD_CapabilitiesDump($s_Comment)
-	If @Compiled Then Return ; because of GDRP reason do not throw nothing to console when compiled script
```
mlipok, line 426: as `__WD_ConsoleWrite` is used and 
```autoit
Global $_WD_DEBUG = ((@compiled) ? ($_WD_DEBUG_None) : ($_WD_DEBUG_Info))
```

this part is not needed

Danp2:
> I suggest that we take a step back and determine which actions would be considered unacceptable under GDPR. Here's a list to start things off 

@mlipok Still looking forward to your feedback / suggestions on this.

FWIW, not everyone will want or need GDPR enforcement. I'm thinking that we should add the ability to optionally adjust UDF functionality for better GDPR. For example --

- Add a global boolean variable $_WD_GDPR_ENFORCE
- _WD_Option would serve as the mechanism to set or retrieve this value
- Add checks for $_WD_GDPR_ENFORCE = True at various points in the UDF where GDPR enforcement is desirable

IMO, this is a better solution than checking @compiled and automatically adjusting defaults.

mlipok:
This change: [$_WD_DEBUG_Info usage > $_WD_DEBUG_Full](https://github.com/Danp2/au3WebDriver/pull/290/commits/2dcda70647571002f4686734ab733bbcb29a3044)
Is somehow related to GDPR as:
```autoit
__WD_ConsoleWrite($sFuncName & ": URL=" & $sURL & "; $sData=" & $sData, $_WD_DEBUG_Info)
```
Can contain sensitive data when you fill up login forms, or fill up some personal data, and thus should be changed to the case when `$_WD_DEBUG_Full` is used (for this reason it was commited).

Danp2:
> This change: [$_WD_DEBUG_Info usage > $_WD_DEBUG_Full](https://github.com/Danp2/au3WebDriver/pull/290/commits/2dcda70647571002f4686734ab733bbcb29a3044)
> Is somehow related to GDPR as:

I would prefer to think that the changes referenced are more in line with #290. GDPR should be addressed separately. Still awaiting your thoughts / feedback on my prior suggestions.

mlipok:
It need's to be closed and should be discussed as issue.

PR #297, opened 2022-04-18
HELPER: $_WD_HTTPRESPONSE = ''

mlipok:
The same change should be done in wd_helper.au3 accordingly.

### DIFF
--- a/wd_helper.au3
+++ b/wd_helper.au3
@@ -317,6 +317,7 @@ Func _WD_WaitElement($sSession, $sStrategy, $sSelector, $iDelay = Default, $iTim
 	Local Const $sFuncName = "_WD_WaitElement"
 	Local $iErr, $sElement, $bIsVisible = True, $bIsEnabled = True
 	$_WD_HTTPRESULT = 0
+	$_WD_HTTPRESPONSE = ''
 
 	If $iDelay = Default Then $iDelay = 0
 	If $iTimeout = Default Then $iTimeout = $_WD_DefaultTimeout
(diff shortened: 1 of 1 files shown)


wd_helper.au3
```diff
@@ -317,6 +317,7 @@ Func _WD_WaitElement($sSession, $sStrategy, $sSelector, $iDelay = Default, $iTim
 	Local Const $sFuncName = "_WD_WaitElement"
 	Local $iErr, $sElement, $bIsVisible = True, $bIsEnabled = True
 	$_WD_HTTPRESULT = 0
+	$_WD_HTTPRESPONSE = ''
```
mlipok, line 320: also what about adding information in function headers ? I mean in remarks.

mlipok: I was thinking about relatedt not remarks.
[_WD_LastHTTPResponse related](https://github.com/Danp2/au3WebDriver/pull/297/commits/14dad47e89c625408399a9d137f1936ed1441f7f)

Danp2: I'm not sure if that is really needed. I think we need to define what qualifies a function as "related".

mlipok: I can do this in opposit way.
I mean to mention each function which change the value, in `_WD_LastHTTPResponse` function header in the related section, but not in any of this function.

In such case we also need to do the same for `_WD_LastHTTPResult`

Danp2: Good idea. I think that would make more sense than adding _WD_LastHTTPResult and _WD_LastHTTPResult as related functions for just about every other WD function. 😄

PR #301, opened 2022-04-23
Add _WD_GetElementByRegEx

TheDcoder:
### Proposed changes

Addition of a new function called `_WD_FindElement_ByRegExp` to match elements with regular expressions.

### Checklist

### Types of changes

- [ ] Other (please describe)

### System under test

N/A - Generic - Universal - Agnostic

### DIFF
--- a/wd_helper.au3
+++ b/wd_helper.au3
@@ -2220,7 +2220,7 @@ EndFunc   ;==>_WD_CheckContext
 ; ===============================================================================================================================
 Func _WD_FindElement_ByRegExp($sSession, $sMode, $sRegEx, $sRegExFlags = "", $bAll = False)
 	Local Static $s_JavaScript = _
-			"return _JS_FindElement_ByRegExp(arguments[0], arguments[1], arguments[2], arguments[3]) || '';" & _
+			"return _JS_FindElement_ByRegExp(arguments[0], arguments[1], arguments[2] , arguments[3]) || '';" & _
 			"function _JS_FindElement_ByRegExp(mode, pattern, flags = '', all = false) {" & _
 			"   var regex = new RegExp(pattern, flags);" & _
 			"   var elements;" & _
@@ -2230,8 +2230,16 @@ Func _WD_FindElement_ByRegExp($sSession, $sMode, $sRegEx, $sRegExFlags = "", $bA
 			""
 
 	Local $sArguments = StringFormat('"%s", "%s", "%s", "%s"', $sMode, $sRegEx, $sRegExFlags, StringLower($bAll))
-	Local $sResult = _WD_ExecuteScript($sSession, $s_JavaScript, $sArguments, False, $_WD_JSON_Value)
-	Return SetError(@error, @extended, $sResult)
+	Local $oValues = _WD_ExecuteScript($sSession, $s_JavaScript, $sArguments, False, $_WD_JSON_Value)
+	If @error Then Return SetError(@error, @extended, $oValues)
+
+	Local $sKey = "[" & $_WD_ELEMENT_ID & "]", $iRow =0
+	Local $aElements[UBound($oValues)]
+	For $oValue In $oValues
+		$aElements[$iRow] = Json_Get($oValue, $sKey)
+		$iRow += 1
+	Next
+	Return SetError(@error, @extended, $aElements)
 EndFunc   ;==>_WD_FindElement_ByRegExp
 
 ; #FUNCTION# ====================================================================================================================
--- a/wd_helper.au3
+++ b/wd_helper.au3
@@ -2228,7 +2228,7 @@ Func _WD_FindElement_ByRegExp($sSession, $sMode, $sRegExPattern, $sRegExFlags = 
 			"}" & _
 			""
 
-	Local $sArguments = StringFormat('"%s", "%s", "%s", "%s"', $sMode, $sRegExPattern, $sRegExFlags, $bAll)
+	Local $sArguments = StringFormat('"%s", "%s", "%s"', $sMode, $sRegExPattern, $sRegExFlags)
 	Local $oValues = _WD_ExecuteScript($sSession, $s_JavaScript, $sArguments, False, $_WD_JSON_Value)
 	If @error Then Return SetError(@error, @extended, $oValues)
 
--- a/wd_helper.au3
+++ b/wd_helper.au3
@@ -2199,34 +2199,35 @@ EndFunc   ;==>_WD_CheckContext
 ; #FUNCTION# ====================================================================================================================
 ; Name ..........: _WD_FindElement_ByRegExp
 ; Description ...: Find element by matching attributes with RegExp
-; Syntax ........: _WD_FindElement_ByRegExp($sSession, $sMode, $sRegEx[, $sRegExFlags = ""])
+; Syntax ........: _WD_FindElement_ByRegExp($sSession, $sMode, $sRegEx[, $sRegExFlags = ""[, $bAll = False]])
 ; Parameters ....: $sSession            - Session ID from _WD_CreateSession
 ;                  $sMode               - Attribute of the element which should be matched, e.g. `id`, `style`, `class` etc.
 ;                  $sRegEx              - RegEx in JavaScript format
 ;                  $sRegExFlags         - [optional] RegEx Flags. Default is "".
+;                  $bAll                - [optional] Fetch a list of all elements that fits to the RegEx pattern ? Default is False - return only first element
 ; Return values .: Success - Desired Element identifier
 ;                  Failure - Response from _WD_ExecuteScript() and sets @error to value returned from _WD_ExecuteScript()
 ; Author ........: TheDcoder
-; Modified ......:
+; Modified ......: mLipok
 ; Remarks .......: The RegEx matching is done by the browser's JavaScript engine so AutoIt's RegEx rules may not accurately work
 ;                  in this function. You may refer to this cheatsheet for further information:
 ;                  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Cheatsheet
 ; Related .......:
 ; Link ..........:
-; Example .......: _WD_FindElement_ByRegExp($sSession, 'class', 'button-[0-9]', 'i')
+; Example .......: _WD_FindElement_ByRegExp($sSession, 'class', 'button-[0-9]', 'i', True)
 ; ===============================================================================================================================
-Func _WD_FindElement_ByRegExp($sSession, $sMode, $sRegEx, $sRegExFlags = "")
+Func _WD_FindElement_ByRegExp($sSession, $sMode, $sRegEx, $sRegExFlags = "", $bAll = False)
 	Local Static $s_JavaScript = _
-			"return _WD_FindElement_ByRegExp(arguments[0], arguments[1], arguments[2]) || "";" & @CRLF & _
-			"function _WD_FindElement_ByRegExp(mode, pattern, flags = "") {" & @CRLF & _
+			"return _WD_FindElement_ByRegExp(arguments[0], arguments[1], arguments[2]) , arguments[3]) || '';" & @CRLF & _
+			"function _WD_FindElement_ByRegExp(mode, pattern, flags = '', all = false) {" & @CRLF & _
 			"   var regex = new RegExp(pattern, flags);" & @CRLF & _
 			"   var elements;" & @CRLF & _
 			"   elements = document.querySelectorAll(`[${mode}]`);" & @CRLF & _
-			"   return Array.prototype.find.call(elements, x => regex.test(x.getAttribute(mode)));" & @CRLF & _
+			"   return Array.prototype[all ? 'filter' : 'find'].call(elements, x => regex.test(x.getAttribute(mode)));" & @CRLF & _
 			"}" & @CRLF & _
 			""
 
-	Local $sArguments = StringFormat('"%s", "%s", "%s"', $sMode, $sRegEx, $sRegExFlags)
+	Local $sArguments = StringFormat('"%s", "%s", "%s, "%s"', $sMode, $sRegEx, $sRegExFlags, StringLower($bAll))
 	Local $sResult = _WD_ExecuteScript($sSession, $s_JavaScript, $sArguments, False, $_WD_JSON_Value)
 	Return SetError(@error, @extended, $sResult)
 EndFunc   ;==>_WD_FindElement_ByRegExp
--- a/wd_helper.au3
+++ b/wd_helper.au3
@@ -2222,33 +2222,36 @@ Func _WD_FindElement_ByRegExp($sSession, $sMode, $sRegExPattern, $sRegExFlags = 
 	Local Const $sFuncName = "_WD_FindElement_ByRegExp"
 	Local $iRow = 0, $iErr = 0, $vResult = ''
 	Local Static $s_JavaScript = _
-			"return _JS_FindElement_ByRegExp(arguments[0], arguments[1], arguments[2]) || '';" & _
-			"function _JS_FindElement_ByRegExp(mode, pattern, flags = '') {" & _
+			"return _JS_FindElement_ByRegExp(arguments[0], arguments[1], arguments[2] , arguments[3]) || '';" & _
+			"function _JS_FindElement_ByRegExp(mode, pattern, flags = '', all = false) {" & _
 			"   var regex = new RegExp(pattern, flags);" & _
-			"   var elements = document.querySelectorAll(`[${mode}]`);" & _
-			"   return Array.prototype['filter'].call(elements, x => regex.test(x.getAttribute(mode)));" & _
+			"   var elements;" & _
+			"   elements = document.querySelectorAll(`[${mode}]`);" & _
+			"   return Array.prototype[all ? 'filter' : 'find'].call(elements, x => regex.test(x.getAttribute(mode)));" & _
 			"}" & _
 			""
 
-	Local $sArguments = StringFormat('"%s", "%s", "%s"', $sMode, $sRegExPattern, $sRegExFlags)
+ 	Local $sArguments = StringFormat('"%s", "%s", "%s", %s', $sMode, $sRegExPattern, $sRegExFlags, StringLower($bAll))
 	Local $oValues = _WD_ExecuteScript($sSession, $s_JavaScript, $sArguments, False, $_WD_JSON_Value)
 	$iErr = @error
 	If Not @error Then
 		Local $sKey = "[" & $_WD_ELEMENT_ID & "]"
-		Local $aElements[UBound($oValues)]
-		If UBound($aElements) < 1 Then
-			$iErr = $_WD_ERROR_NoMatch
-		Else
-			For $oValue In $oValues
-				$aElements[$iRow] = Json_Get($oValue, $sKey)
-				$iRow += 1
-			Next
 
-			If $bAll Then
-				$vResult = $aElements
+		If $bAll Then
+			Local $aElements[UBound($oValues)]
+			If UBound($aElements) < 1 Then
+				$iErr = $_WD_ERROR_NoMatch
 			Else
-				$vResult = $aElements[0]
+				For $oValue In $oValues
+					$aElements[$iRow] = Json_Get($oValue, $sKey)
+					$iRow += 1
+				Next
+
+				$vResult = $aElements
 			EndIf
+		Else
+			$vResult = Json_Get($oValues, $sKey)
+			$iRow = 1
 		EndIf
 	EndIf
 
--- a/wd_helper.au3
+++ b/wd_helper.au3
@@ -2198,14 +2198,14 @@ EndFunc   ;==>_WD_CheckContext
 
 ; #FUNCTION# ====================================================================================================================
 ; Name ..........: _WD_FindElement_ByRegExp
-; Description ...: Find element identifier by matching attributes with RegExp
+; Description ...: Find element identifier by matching attributes values using regular expression
 ; Syntax ........: _WD_FindElement_ByRegExp($sSession, $sMode, $sRegEx[, $sRegExFlags = ""[, $bAll = False]])
 ; Parameters ....: $sSession            - Session ID from _WD_CreateSession
 ;                  $sMode               - Attribute of the element which should be matched, e.g. `id`, `style`, `class` etc.
 ;                  $sRegEx              - RegEx in JavaScript format
 ;                  $sRegExFlags         - [optional] RegEx Flags. Default is "".
 ;                  $bAll                - [optional] Fetch a list of all elements that fits to the RegEx pattern ? Default is False - return only first element
-; Return values .: Success - Desired Element identifier
+; Return values .: Success - matched attribute value
 ;                  Failure - Response from _WD_ExecuteScript() and sets @error to value returned from _WD_ExecuteScript()
 ; Author ........: TheDcoder
 ; Modified ......: mLipok
--- a/wd_helper.au3
+++ b/wd_helper.au3
@@ -2213,6 +2213,7 @@ EndFunc   ;==>_WD_CheckContext
 ;                  in this function. You may refer to this cheatsheet for further information:
 ;                  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Cheatsheet
 ;                  The matching elements are returned in the same format as the results from _WD_FindElement
+;                  As JavaScript's RegEx engine differs from AutoIt's PCRE engine you can use https://regex101.com with FLAVOR set to: ECMAScript (JavaScript) to validate your RegEx
 ; Related .......:
 ; Link ..........:
 ; Example .......: _WD_FindElement_ByRegExp($sSession, 'class', 'button-[0-9]', 'i', True)
--- a/wd_helper.au3
+++ b/wd_helper.au3
@@ -2197,6 +2197,41 @@ Func _WD_CheckContext($sSession, $bReconnect = Default, $vTarget = Default)
 EndFunc   ;==>_WD_CheckContext
 
 ; #FUNCTION# ====================================================================================================================
+; Name ..........: _WD_FindElement_ByRegExp
+; Description ...: Find element by matching attributes with RegExp
+; Syntax ........: _WD_FindElement_ByRegExp($sSession, $sMode, $sRegEx[, $sRegExFlags = ""])
+; Parameters ....: $sSession            - Session ID from _WD_CreateSession
+;                  $sMode               - Attribute of the element which should be matched, e.g. `id`, `style`, `class` etc.
+;                  $sRegEx              - RegEx in JavaScript format
+;                  $sRegExFlags         - [optional] RegEx Flags. Default is "".
+; Return values .: Success - Desired Element identifier
+;                  Failure - Response from _WD_ExecuteScript() and sets @error to value returned from _WD_ExecuteScript()
+; Author ........: TheDcoder
+; Modified ......:
+; Remarks .......: The RegEx matching is done by the browser's JavaScript engine so AutoIt's RegEx rules may not accurately work
+;                  in this function. You may refer to this cheatsheet for further information:
+;                  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Cheatsheet
+; Related .......:
+; Link ..........:
+; Example .......: _WD_FindElement_ByRegExp($sSession, 'class', 'button-[0-9]', 'i')
+; ===============================================================================================================================
+Func _WD_FindElement_ByRegExp($sSession, $sMode, $sRegEx, $sRegExFlags = "")
+	Local Static $s_JavaScript = _
+			"return _WD_FindElement_ByRegExp(arguments[0], arguments[1], arguments[2]) || "";" & @CRLF & _
+			"function _WD_FindElement_ByRegExp(mode, pattern, flags = "") {" & @CRLF & _
+			"   var regex = new RegExp(pattern, flags);" & @CRLF & _
+			"   var elements;" & @CRLF & _
+			"   elements = document.querySelectorAll(`[${mode}]`);" & @CRLF & _
+			"   return Array.prototype.find.call(elements, x => regex.test(x.getAttribute(mode)));" & @CRLF & _
+			"}" & @CRLF & _
+			""
+
+	Local $sArguments = StringFormat('"%s", "%s", "%s"', $sMode, $sRegEx, $sRegExFlags)
+	Local $sResult = _WD_ExecuteScript($sSession, $s_JavaScript, $sArguments, False, $_WD_JSON_Value)
+	Return SetError(@error, @extended, $sResult)
+EndFunc   ;==>_WD_FindElement_ByRegExp
+
+; #FUNCTION# ====================================================================================================================
 ; Name ..........: _WD_JsonActionKey
 ; Description ...: Formats keyboard "action" strings for use in _WD_Action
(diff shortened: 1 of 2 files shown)


mlipok:
+1

Danp2:
Thanks for the submission. A few questions / comments for you --

- Does this return an array when multiple matches exist? 
- You should add `__WD_Error` to your Return statement for consistent logging
- Renaming this to `_WD_GetElementByRegExp` or something similar would work better IMO

mlipok:
I will take a look in the comming weekend.

mlipok:
> * Does this return an array when multiple matches exist?

Take a look on:
[Fetch a list of all elements](https://github.com/Danp2/au3WebDriver/pull/301/commits/bfd5035e78d70bd062f8b9e08f742a617baef6dd)

This is WIP not checked yet.

mlipok:
for example I have such element
```html
<a class="edit" id="ctl00_ContentPlaceHolder1_klauzula_8441556_AddExecutionButton" href="javascript:__doPostBack('ctl00$ContentPlaceHolder1$klauzula_8441556$AddExecutionButton','')">SOME TEXT</a>
```
It is not always `ctl00$ContentPlaceHolder1$klauzula_8441556$AddExecutionButton` sometimes it is `ctl00$ContentPlaceHolder1$nakaz_11231433$AddExecutionButton`
My usage is:
```autoit

_WD_FindElement_ByRegExp($_ESAD__WD_SESSION, 'id', '(?i)ctl00_ContentPlaceHolder1_[klauzulanakaz]*?_\d+_AddExecutionButton', 'i')
```
and it returns:
string that represents value of ID attribute `ctl00$ContentPlaceHolder1$klauzula_8441556$AddExecutionButton` or `ctl00$ContentPlaceHolder1$nakaz_11231433$AddExecutionButton`

mlipok:
> My usage is:
> 
> ```autoit
> _WD_FindElement_ByRegExp($_ESAD__WD_SESSION, 'id', '(?i)ctl00_ContentPlaceHolder1_[klauzulanakaz]*?_\d+_AddExecutionButton', 'i')
> ```

the code was WIP status as we have doing migration form IE to WD and as so far we do all our changes in the background without checking !!!

So today after
[matched element description fix](https://github.com/Danp2/au3WebDriver/pull/301/commits/65eb6e0a0660198b4da7c39803ce12505dc67056)

i made a correction to my code
```autoit
Func _WD_FindElementID_ByRegExp($sSession, $sRegEx, $sRegExFlags = "")
	Local $sElement = _WD_FindElement_ByRegExp($sSession, 'id', $sRegEx, $sRegExFlags, False)
	If @error Then Return SetError(@error, @extended, $sElement)

	Local $sElement_ID = _WD_ElementAction($sSession, $sElement, 'ATTRIBUTE', 'id')
	Return SetError(@error, @extended, $sElement_ID )
EndFunc   ;==>_WD_FindElementID_ByRegExp
```

mlipok:
> the code was WIP status as we have doing migration form IE to WD and as so far we do all our changes in the background without checking !!!

In the comming weekend I need to do all the test so the WIP status will be verified.

mlipok:
Should this sufix be `_ByRegExp` or rather _ByRegEx` I mean with `p` or without `p` on the end.

TheDcoder:
@mlipok I think we should use the same scheme as the official AutoIt functions.

mlipok:
So as I understand you say that the current name  **_WD_FindElement_ByRegExp** is fine as we have
- String**RegExp**()
- String**RegExp**Replace()

TheDcoder:
Yes. I think that was my motivation when I decided to use this name, but I don't really remember now.

wd_helper.au3
```diff
@@ -2196,6 +2196,43 @@ Func _WD_CheckContext($sSession, $bReconnect = Default, $vTarget = Default)
...
+;                  $sRegExFlags         - [optional] RegEx Flags. Default is "".
+;                  $bAll                - [optional] Fetch a list of all elements that fits to the RegEx pattern ? Default is False - return only first element
+; Return values .: Success - Matched element
+;                  Failure - Response from _WD_ExecuteScript() and sets @error to value returned from _WD_ExecuteScript()
+; Author ........: TheDcoder
+; Modified ......: mLipok
+; Remarks .......: The RegEx matching is done by the browser's JavaScript engine so AutoIt's RegEx rules may not accurately work
+;                  in this function. You may refer to this cheatsheet for further information:
+;                  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Cheatsheet
+;                  The returned matched elements are in the same format like result form _WD_FindElement and _WD_WaitElement
```
mlipok, line 2215: `like result` or `like results` ?

Danp2: `results`. Also `form` should be `from`. I would propose to change this sentence to something closer to this--

`The matching elements are returned in the same format as the results from _WD_FindElement`

mlipok: O

mlipok: Please take a look:
[description rewording](https://github.com/Danp2/au3WebDriver/pull/301/commits/6974f29d5e2769255c9a52e7eee83b93d8730047)

Danp2 (changes requested):
Still waiting on responsed to the following items 😉 --

    Does this return an array when multiple matches exist?
    You should add __WD_Error to your Return statement for consistent logging
    Renaming this to _WD_GetElementByRegExp or something similar would work better IMO

wd_helper.au3
```diff
@@ -2196,6 +2196,44 @@ Func _WD_CheckContext($sSession, $bReconnect = Default, $vTarget = Default)
...
+;                  $sRegExFlags         - [optional] RegEx Flags. Default is "".
+;                  $bAll                - [optional] Fetch a list of all elements that fits to the RegEx pattern ? Default is False - return only first element
+; Return values .: Success - Matched element
+;                  Failure - Response from _WD_ExecuteScript() and sets @error to value returned from _WD_ExecuteScript()
+; Author ........: TheDcoder
+; Modified ......: mLipok
+; Remarks .......: The RegEx matching is done by the browser's JavaScript engine so AutoIt's RegEx rules may not accurately work
+;                  in this function. You may refer to this cheatsheet for further information:
+;                  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Cheatsheet
+;                  The matching elements are returned in the same format as the results from _WD_FindElement
```
Danp2, line 2215: This is not currently how the function is operating. The resulting array contains objects instead of strings. The coding to handle this is already in place in _WD_FindElement and would need to be replicated here or moved to an internal function.

Danp2 (changes requested):

wd_helper.au3
```diff
@@ -2199,34 +2199,35 @@ EndFunc   ;==>_WD_CheckContext
...
 			"   var regex = new RegExp(pattern, flags);" & @CRLF & _
 			"   var elements;" & @CRLF & _
 			"   elements = document.querySelectorAll(`[${mode}]`);" & @CRLF & _
-			"   return Array.prototype.find.call(elements, x => regex.test(x.getAttribute(mode)));" & @CRLF & _
+			"   return Array.prototype[all ? 'filter' : 'find'].call(elements, x => regex.test(x.getAttribute(mode)));" & @CRLF & _
 			"}" & @CRLF & _
 			""
 
-	Local $sArguments = StringFormat('"%s", "%s", "%s"', $sMode, $sRegEx, $sRegExFlags)
+	Local $sArguments = StringFormat('"%s", "%s", "%s, "%s"', $sMode, $sRegEx, $sRegExFlags, StringLower($bAll))
```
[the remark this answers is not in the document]
mlipok: Yes it looks good.
Thanks for fixing.

mlipok:
> Still waiting on responsed to the following items 😉 --
> 
> ```
> Does this return an array when multiple matches exist?
> You should add __WD_Error to your Return statement for consistent logging
> Renaming this to _WD_GetElementByRegExp or something similar would work better IMO
> ```

I just now taking a look and going to make an example script

wd_helper.au3
mlipok: Please take a look on: [returning as array](https://github.com/Danp2/au3WebDriver/pull/301/commits/b3a4be35cf5d87467ed5949c2960ee06d2a595c4)

mlipok: my testing script:
```autoit
Func _EXAMPLE__WD_FindElement_ByRegExp($sSesion)
	_WD_Navigate($sSesion, 'https://github.com/Danp2/au3WebDriver/commits/master')
	_WD_LoadWait($sSesion)
;~ 	data-url="/Danp2/au3WebDriver/commits/8bdf053f3b97d0b585b9f5bf2d6607a67eb5a545/commits_list_item"
	Local $vResult = _WD_FindElement_ByRegExp($sSesion, 'data-url', '\/danp2.au3webdr.*\/commits_list_item', 'i', False)
	If @error Then ConsoleWrite("! ---> @error=" & @error & "  @extended=" & @extended & _
			" : _WD_FindElement_ByRegExp" & @CRLF)

	_ArrayDisplay($vResult, '$vResult')
EndFunc   ;==>_EXAMPLE__WD_FindElement_ByRegExp
```

result:
image

wd_helper.au3
```diff
@@ -2196,6 +2196,52 @@ Func _WD_CheckContext($sSession, $bReconnect = Default, $vTarget = Default)
...
+; Remarks .......: The RegEx matching is done by the browser's JavaScript engine so AutoIt's RegEx rules may not accurately work
+;                  in this function. You may refer to this cheatsheet for further information:
+;                  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Cheatsheet
+;                  The matching elements are returned in the same format as the results from _WD_FindElement
+;                  As JavaScript's RegEx engine differs from AutoIt's PCRE engine you can use https://regex101.com with FLAVOR set to: ECMAScript (JavaScript) to validate your RegEx
+; Related .......:
+; Link ..........:
+; Example .......: _WD_FindElement_ByRegExp($sSession, 'class', 'button-[0-9]', 'i', True)
+; ===============================================================================================================================
+Func _WD_FindElement_ByRegExp($sSession, $sMode, $sRegEx, $sRegExFlags = "", $bAll = False)
```
mlipok, line 2221: using `$bAll = False` like this:
```autoit
_WD_FindElement_ByRegExp($sSesion, 'data-url', '\/danp2.au3webdr.*\/commits_list_item', 'i', False)
```
it still returns all 35 elements 
should only return first as string not as array

mlipok (changes requested):

wd_helper.au3
mlipok: it requires: [returning as array](https://github.com/Danp2/au3WebDriver/pull/301/commits/b3a4be35cf5d87467ed5949c2960ee06d2a595c4)

wd_helper.au3
mlipok: This fix [return string or array fix](https://github.com/Danp2/au3WebDriver/pull/301/commits/6ab8649f0f227dfc82d5b7bd730318e47a9a3eaf)

When I use this example:

```autoit
Func _EXAMPLE__WD_FindElement_ByRegExp($sSesion)
	_WD_Navigate($sSesion, 'https://github.com/Danp2/au3WebDriver/commits/master')
	_WD_LoadWait($sSesion)
;~ 	data-url="/Danp2/au3WebDriver/commits/8bdf053f3b97d0b585b9f5bf2d6607a67eb5a545/commits_list_item"

	Local $vResult = _WD_FindElement_ByRegExp($sSesion, 'data-url', '\/danp2.au3webdr.*\/commits_list_item', 'i', False)
	If @error Then ConsoleWrite("! ---> @error=" & @error & "  @extended=" & @extended & _
			" : _WD_FindElement_ByRegExp" & @CRLF)

	MsgBox($MB_OK + $MB_TOPMOST + $MB_ICONINFORMATION, "Information", $vResult)

	Local $vResult = _WD_FindElement_ByRegExp($sSesion, 'data-url', '\/danp2.au3webdr.*\/commits_list_item', 'i', True)
	If @error Then ConsoleWrite("! ---> @error=" & @error & "  @extended=" & @extended & _
			" : _WD_FindElement_ByRegExp" & @CRLF)

	_ArrayDisplay($vResult, '$vResult')
EndFunc   ;==>_EXAMPLE__WD_FindElement_ByRegExp
```

gives me such results:
image


What you think about ?

Danp2: You were converting all Javascript parameters to strings. The correct fix IMO is to change this one line --

`Local $sArguments = StringFormat('"%s", "%s", "%s", %s', $sMode, $sRegEx, $sRegExFlags, StringLower($bAll))`

Notice that the last `%s` is no longer surrounded by double quotes. Now that variable will be seen as a boolean instead of a string.

mlipok: Thanks this should do the trick.
But I want to keep my current version for getting know if there was single or multiple elements returned by the given `$sRegExPattern`.
For this reason I add `$iRow` to be returned as `@extended`

mlipok: Please check my current [refactoring + standardarization + error handling](https://github.com/Danp2/au3WebDriver/pull/301/commits/3fdfa3f12e4615e45273058102b9f34672e8fadc)


```autoit
Func _EXAMPLE__WD_FindElement_ByRegExp($sSesion)
	_WD_Navigate($sSesion, 'https://github.com/Danp2/au3WebDriver/commits/master')
	_WD_LoadWait($sSesion)
;~ 	data-url="/Danp2/au3WebDriver/commits/8bdf053f3b97d0b585b9f5bf2d6607a67eb5a545/commits_list_item"

	Local $vResult, $sRegExpPattern = ''


	$sRegExpPattern = '\/danp2.au3webdr.*\/commits_list_item'
	$vResult = _WD_FindElement_ByRegExp($sSesion, 'data-url', $sRegExpPattern, 'i', False)
	MsgBox($MB_OK + $MB_TOPMOST, "Information", _
			"$vResult = " & $vResult & @CRLF & _
			"" & @CRLF & _
			"@error = " & @error & @CRLF & _
			"@extended = " & @extended & @CRLF & _
			"")

	$sRegExpPattern = '\/danp2.*\/commits_list_item'
	$vResult = _WD_FindElement_ByRegExp($sSesion, 'data-url', $sRegExpPattern, 'i', True)
	If @error Then ConsoleWrite("! ---> @error=" & @error & "  @extended=" & @extended & _
			" : _WD_FindElement_ByRegExp" & @CRLF)

	_ArrayDisplay($vResult, '$vResult')
EndFunc   ;==>_EXAMPLE__WD_FindElement_ByRegExp
```
It gives me:
image

TheDcoder: > Thanks this should do the trick. But I want to keep my current version for getting know if there was single or multiple elements returned by the given `$sRegExPattern`. For this reason I add `$iRow` to be returned as `@extended`

In my opinion we should not alter this code because it will impact performance for cases where we don't need to know the number of matches. This can still be accomplished by the calling code by setting `$bAll` to true and counting the elements manually.

The number of matches can be a lot if a common `mode` attribute like class is chosen and the performance impact can be significant as the code will have to go through each element even when `$bAll` is not set.

In sort, I think we are over accommodating this function by sacrificing performance, just my two cents :smile:

Danp2: I agree with @TheDcoder. The Javascript code should be reverted to the earlier incarnation and then fix the parameter passing so that the $bAll functionality works properly. @mlipok Unless I misunderstood your comment, you should still be able to get the count of matching elements.

mlipok: Ok.
Yes.

Danp2:
> Func _WD_FindElementName_ByRegExp
> Func _WD_FindElementID_ByRegExp

@mlipok Curious about your use/case for these? What do you intend to do with the name or ID once you retrieve it?

mlipok:
Wondering why this following JS works
`\/danp2.*[0-9]+\/commits_list_item`

but not this one
`\/danp2.*\d+\/commits_list_item`

Danp2:
Not sure, but I don't believe either xpath is correct since a commit hash can also contain some alpha characters. You probably need to check for `[0-9a-f]`

mlipok:
> You probably need to check for `[0-9a-f]`

already checked `\/danp2.*[0-9]+\/commits_list_item` and as I said it works.

mlipok:
> > Func _WD_FindElementName_ByRegExp
> > Func _WD_FindElementID_ByRegExp
> 
> @mlipok Curious about your use/case for these? What do you intend to do with the name or ID once you retrieve it?

for compability with some project that I have in InternetExplorer.
I need them to make "easy" migration.

Danp2:
> already checked `\/danp2.*[0-9]+\/commits_list_item` and as I said it works.

It may have returned a result, but they were incomplete. Try it with the new option I posted and you'll see that more results are returned.

wd_helper.au3
```diff
@@ -2213,6 +2213,7 @@ EndFunc   ;==>_WD_CheckContext
 ;                  in this function. You may refer to this cheatsheet for further information:
 ;                  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Cheatsheet
 ;                  The matching elements are returned in the same format as the results from _WD_FindElement
+;                  As JavaScript's RegEx engine differs from AutoIt's PCRE engine you can use https://regex101.com with FLAVOR set to: ECMAScript (JavaScript) to validate your RegEx
```
Danp2, line 2216: This remark is should be either eliminated or combined with the existing remarks that already attempt to explain that Autoit RegEx <> Javascript RegEx

Danp2 (changes requested):

wd_helper.au3
Danp2: See a7383967fc813f3012b7a6b1739ae82aa7836506 for my attempt to address this.

wd_helper.au3
```diff
@@ -2196,6 +2196,68 @@ Func _WD_CheckContext($sSession, $bReconnect = Default, $vTarget = Default)
 	Return SetError(__WD_Error($sFuncName, ($iResult) ? $_WD_ERROR_Success : $_WD_ERROR_Exception), 0, $iResult)
 EndFunc   ;==>_WD_CheckContext
 
+; #FUNCTION# ====================================================================================================================
+; Name ..........: _WD_FindElement_ByRegExp
```
Danp2, line 2200: I propose that this function should be renamed for the following reasons --

- None of the existing functions utilize an underscore in their name other than the initial `_WD_`
- Other "helper" functions use the naming convention `_WD_GetElementByxxx`

Some possible alternatives --

- `_WD_GetElementByRegEx`, previously suggested without any feedback received. 
- `_WD_FindElementEx`, indicating that this function is an advanced or extended version of `_WD_FindElement`.
- `_WD_GetElementByJSSelector`

Danp2 (changes requested):

wd_helper.au3
mlipok: maybe `_WD_GetElementByJSRegExp` ?

wd_helper.au3
mlipok: I see no obstacles, so please propose the final version.

mlipok:
> > already checked `\/danp2.*[0-9]+\/commits_list_item` and as I said it works.
> 
> It may have returned a result, but they were incomplete. Try it with the new option I posted and you'll see that more results are returned.

checking

wd_helper.au3
mlipok: Please take a look on: [_WD_FindElement_ByRegExp > _WD_GetElementByRegEx](https://github.com/Danp2/au3WebDriver/pull/301/commits/281028b9fbb69b81d0a99838059447a728ac09bb)

wd_helper.au3
mlipok: > See [a738396](https://github.com/Danp2/au3WebDriver/commit/a7383967fc813f3012b7a6b1739ae82aa7836506) for my attempt to address this.

I see you have reverte this by: [Revert to earlier version of Javascript](https://github.com/Danp2/au3WebDriver/pull/301/commits/a7383967fc813f3012b7a6b1739ae82aa7836506)

Ok.

mlipok:
I incorporate this change: [JS refactoring to not use arguments](https://github.com/Danp2/au3WebDriver/pull/301/commits/05078644b7488097ee5790397b3fb2e555247ec9)

Because as a result in console you get:

mlipok:
btw.
This following usage:
```autoit
	$sRegExpPattern = '/danp2.*\d+/commits_list_item$'
	$vResult = _WD_GetElementByRegEx($sSesion, 'data-url', $sRegExpPattern, 'i', True)
```
did not work with WebDriver.
I get in console:
```
1653514246004   webdriver::server       DEBUG   -> POST /session/620caa86-56f8-4334-a5d1-01722329cc9b/execute/sync {"script":"return _JS_FindElementByRegExp('data-url', '/danp2.*\\d+/commits_list_item$', 'i', true) || '';function _JS_FindElementByRegExp(mode, pattern, flags = '', all = false) {   var regex = new RegExp(pattern, flags);   var elements;   elements = document.querySelectorAll(`[${mode}]`);   return Array.prototype[all ? 'filter' : 'find'].call(elements, x => regex.test(x.getAttribute(mode)));}", "args":[]}
1653514246039   webdriver::server       DEBUG   <- 200 OK {"value":[]}
```

but using this:
```js
return _JS_FindElementByRegExp('data-url', '/danp2.*\\d+/commits_list_item$', 'i', true) || '';function _JS_FindElementByRegExp(mode, pattern, flags = '', all = false) {   var regex = new RegExp(pattern, flags);   var elements;   elements = document.querySelectorAll(`[${mode}]`);   return Array.prototype[all ? 'filter' : 'find'].call(elements, x => regex.test(x.getAttribute(mode)));}
```
in `Browser console` I get:
image

mlipok:
As I said before this following did not work:
```autoit
$sRegExpPattern = '/danp2.*\d+/commits_list_item$'
$vResult = _WD_GetElementByRegEx($sSesion, 'data-url', $sRegExpPattern, 'i', True)
```

this following works well:
```autoit
$sRegExpPattern = '/danp2.*[0-9]+/commits_list_item$'
$vResult = _WD_GetElementByRegEx($sSesion, 'data-url', $sRegExpPattern, 'i', True)
```

Danp2:
> \d (fails) vs [0-9] (works)

I suspect that this issue is due to the escaping of the backslash character. I tried disabling the escape mechanism, but then I got this --

```
1653578054710   webdriver::server       DEBUG   -> POST /session/c1104011-b051-4beb-a899-60a651b38ea0/execute/sync {"script":"return _JS_FindElementByRegExp('data-url', '/danp2.*\d+/commits_list_item$', 'i', true) || '';function _JS_FindElementByRegExp(mode, pattern, flags = '', all = false) {   var regex = new RegExp(pattern, flags);   var elements;   elements = document.querySelectorAll(`[${mode}]`);   return Array.prototype[all ? 'filter' : 'find'].call(elements, x => regex.test(x.getAttribute(mode)));}", "args":[]}
1653578054710   webdriver::server       DEBUG   <- 400 Bad Request {"value":{"error":"invalid argument","message":"Failed to decode request as JSON: {\"script\":\"return _JS_FindElementByRegExp('data-url', '/danp2.*\\d+/commits_list_item$', 'i', true) || '';function _JS_FindElementByRegExp(mode, pattern, flags = '', all = false) {   var regex = new RegExp(pattern, flags);   var elements;   elements = document.querySelectorAll(`[${mode}]`);   return Array.prototype[all ? 'filter' : 'find'].call(elements, x => regex.test(x.getAttribute(mode)));}\", \"args\":[]}","stacktrace":"Syntax error at :1:65"}}
```

wd_helper.au3
Danp2: See 1794062127aff5a41757431ef3ea1161d748c7be for my revisions

wd_helper.au3
Danp2: That's fine with me. I've updated the function header in an attempt to better clarify the use of Javascript. Please review when you have a moment.

Danp2:
> '/danp2.*\d+/commits_list_item$'

This gets passed to the webdriver as `'/danp2.*\\d+/commits_list_item$'`, but by the time it is process by Javascript it looks like `/danp2.*d+/commits_list_item$`, which is why it is failing.

When I change the regex to `'/danp2.*\\d+/commits_list_item$'`, it gets passed to the webdriver as `/danp2.*\\\\d+/commits_list_item$`, and Javascript sees it as `/danp2.*\d+/commits_list_item$`, and the search is successful.

This can be resolved by adding this line to the function --

` 	sRegExPattern = StringReplace(sRegExPattern, '\', '\\')`

mlipok:
Nicely done: 
[_WD_GetElementByRegEx: Revise function header](https://github.com/Danp2/au3WebDriver/pull/301/commits/1794062127aff5a41757431ef3ea1161d748c7be)

Danp2:
> '/danp2.*\d+/commits_list_item$'

This regex doesn't work in regex101.com because of:

- unescaped forward slashes
- $ symbol

I understand the issue with forward slashes. Why are you including the dollar sign in this regex?

mlipok:
> Why are you including the dollar sign in this regex?

`$ asserts position at the end of a line`

This following example will explain what I'm checking for
https://regex101.com/r/bASVYk/1

I have similar scenario in my cases.

mlipok:
> This can be resolved by adding this line to the function --
> 
> ` sRegExPattern = StringReplace(sRegExPattern, '\', '\\')`

Thanks again for this fix.

Please take a look:
[$sRegExPattern backslashes fix](https://github.com/Danp2/au3WebDriver/pull/301/commits/bb90b2cd074653df9f83bd4bb46cd443a66c7591)

TheDcoder:
I've reviewed the final code, LGTM :)

I've requested the final review from @Danp2 before the merge! Thanks to @mlipok for refining the function and helping it get upstreamed :+1:

Danp2:
@TheDcoder I fixed one issue [here](https://github.com/Danp2/au3WebDriver/pull/301/commits/02179b93b6830297475a37c482ca212c6d5a0522). The one remaining item I see has to do with return error codes --

- `$_WD_ERROR_NoMatch` isn't documented in function header
- I wonder if it would make more sense to return a single error code (ie: $_WD_ERROR_InvalidExpression) instead of whatever is returned by _WD_ExecuteScript

TheDcoder:
@Danp2 Thanks for fixing the bug.

> I wonder if it would make more sense to return a single error code (ie: $_WD_ERROR_InvalidExpression) instead of whatever is returned by _WD_ExecuteScript

I don't think this makes sense because "failure" can only occur when the request isn't processed properly, most RegEx will be valid even though it might not be what the user wants, it's hard to make an invalid expression regex involuntarily so I don't think we need to have a return code for this.

So realistically the failure can only happen in two cases:
1. No match, we are already returning the no match error code here if I recall correctly
2. Something else, i.e an error from ExecuteScript :)

Danp2:
> So realistically the failure can only happen in two cases:
>    No match
>   Something else, i.e an error from ExecuteScript :)

Ok, but is the error code from _WD_ExecuteScript meaningful in the context of _WD_GetElementByRegEx? For example, this is the output I received when supplying an invalid regex flag --

```
__WD_Post: URL=HTTP://127.0.0.1:4444/session/46c0591f-dbb4-4728-9ee7-a003c03f3627/execute/sync; $sData={"script":"return _JS_GetElementByRegEx('data-url', '/danp2.*\\\\d+/commits_list_item$', 'aaai', true) || '';function _JS_GetElementByRegEx(mode, pattern, flags = '', all = false) {   var regex = new RegExp(pattern, flags);   var elements;   elements = document.querySelectorAll(`[${mode}]`);   return Array.prototype[all ? 'filter' : 'find'].call(elements, x => regex.test(x.getAttribute(mode)));}", "args":[]}
__WD_Post ==> Webdriver Exception (10) HTTP status = 500 : ResponseText={"value":{"error":"javascript error","message":"SyntaxError: invalid regular expression flag a","sta...
_WD_ExecuteScript ==> Webdriver Exception (10) HTTP status = 500 :  : Response={"value":{"error":"javascript error","message":"SyntaxError: invalid regular expression flag a","sta...
_WD_GetElementByRegEx ==> Webdriver Exception (10) HTTP status = 500
```

Are there other return codes that might better describe the scenario than  `Webdriver Exception`?

TheDcoder:
@Danp2 Oops, didn't realize that this could happen. But there is an easy fix, we can simply catch the error in JavaScript and return no match :smile:

I'll work with @mlipok to get a fix implemented.

mlipok:
checking

TheDcoder:
@mlipok Can you replace the outer return statement in JavaScript with this?
```javascript
try {
	return _JS_GetElementByRegEx('%s', '%s', '%s', %s) || '';
} catch (e) {
	return '';
}
```

It will simply return an empty string if the `RegEx` constructor throws an error.

Danp2:
@TheDcoder That would work, but the end result is that the user will be told there wasn't a match. An alternative would be to make the change in __WD_DetectError, so that an error still occurs but one more appropriate than `Webdriver Exception`.

TheDcoder:
@Danp2 That's technically true because an invalid RegEx doesn't match anything :wink:

In any case, I'm not familiar with `__WD_DetectError`, so maybe you or mLipok can look into this, I have no objections for such a change. You're the best people for this kind of stuff anyway :)

mlipok:
> An alternative would be to make the change in __WD_DetectError, so that an error still occurs but one more appropriate than `Webdriver Exception`.

Please provide this fix.

Danp2:
__WD_DetectError updated so that "javascript error" now translates to $_WD_ERROR_InvalidExpression. If desired, this can be changed if you don't like using that error code. We could add a new code for this specific error, but I was trying to avoid that if possible.

Here's the updated log --
```
__WD_Post: URL=HTTP://127.0.0.1:4444/session/74e92f63-1e7a-491a-a200-26a3d6947aaf/execute/sync; $sData={"script":"return _JS_GetElementByRegEx('data-url', '/danp2.*\\\\d+/commits_list_item$', 'aaai', true) || '';function _JS_GetElementByRegEx(mode, pattern, flags = '', all = false) {   var regex = new RegExp(pattern, flags);   var elements;   elements = document.querySelectorAll(`[${mode}]`);   return Array.prototype[all ? 'filter' : 'find'].call(elements, x => regex.test(x.getAttribute(mode)));}", "args":[]}
__WD_Post ==> Invalid Expression (11) HTTP status = 500 : ResponseText={"value":{"error":"javascript error","message":"SyntaxError: invalid regular expression flag a","sta...
_WD_ExecuteScript ==> Invalid Expression (11) HTTP status = 500 :  : Response={"value":{"error":"javascript error","message":"SyntaxError: invalid regular expression flag a","sta...
_WD_GetElementByRegEx ==> Invalid Expression (11) HTTP status = 500
```

mlipok:
according to: `Response={"value":{"error":"javascript error",`  
please change `$_WD_ERROR_InvalidExpression` >>> `$_WD_ERROR_JavaScript`

Danp2:
>  please change this name: $_WD_ERROR_InvalidExpression >>> $_WD_ERROR_JavaScript

@mlipok So you are saying that you want to add a new error code to represent the javascript error?

mlipok:
> @mlipok So you are saying that you want to add a new error code to represent the javascript error?

Exactly.

mlipok:
but..... there can be cases when `$_WD_ERROR_InvalidExpression` will fit better.

For example with this response: `Response={"value":{"error":"javascript error","message":"SyntaxError: invalid regular expression flag a`
I mean this is `"error":"javascript error"` and at first glance it is `$_WD_ERROR_JavaScript` 
but message contains `invalid regular expression` so this should be `$_WD_ERROR_InvalidExpression`

but in case when there is no expression error then it should report $_WD_ERROR_JavaScript


But isn't this a broader change (beyond the reach of this PR) and doesn't it require a separate PR?

Danp2 (approved):

PR #320, opened 2022-05-30
Examples

mlipok:
### Proposed changes

https://github.com/Danp2/au3WebDriver/issues/133

### Checklist

- [X] I have read and noticed the [CODE OF CONDUCT](https://github.com/Danp2/au3WebDriver/blob/master/docs/CODE_OF_CONDUCT.md) document
- [X] I have read and noticed the [CONTRIBUTING](https://github.com/Danp2/au3WebDriver/blob/master/docs/CONTRIBUTING.md) document
- [ ] I have added necessary documentation or screenshots (if appropriate)

### Types of changes

Please check `x` the type of change your PR introduces:

- [ ] Bugfix (change which fixes an issue)
- [ ] Feature (change which adds functionality)
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (functional, structural)
- [ ] Documentation content changes
- [x] Other (please describe)

### What is the current behavior?

https://github.com/Danp2/au3WebDriver/issues/133

### What is the new behavior?

Strarting with examples

### Additional context

https://github.com/Danp2/au3WebDriver/issues/133

### System under test

Not related

PR #290, opened 2022-04-10
__WD_Error() minimizing the amount of messages

mlipok:
### Proposed changes

Describe the big picture of your changes here to communicate to the maintainers why we should accept this pull request.<br>
Please ensure you have read and noticed the checklist below.

### Checklist

Put an `x` in the boxes that apply. If you're unsure about any of them, don't hesitate to ask. We are here to help!<br>
This is simply a reminder of what we are going to look for before merging your code.

- [x] I have read and noticed the [CODE OF CONDUCT](https://github.com/Danp2/au3WebDriver/blob/master/docs/CODE_OF_CONDUCT.md) document
- [x] I have read and noticed the [CONTRIBUTING](https://github.com/Danp2/au3WebDriver/blob/master/docs/CONTRIBUTING.md) document
- [ ] I have added necessary documentation or screenshots (if appropriate)

### Types of changes

Please check `x` the type of change your PR introduces:

- [ ] Bugfix (change which fixes an issue)
- [ ] Feature (change which adds functionality)
- [ ] Code style update (formatting, renaming)
- [x] Refactoring (functional, structural)
- [ ] Documentation content changes
- [ ] Other (please describe)

### What is the current behavior?

1.
Currently UDF throws to many messages to console.

2.
Very often they are duplicated. 
for example $sResponseText is showed by `__WD_Delete` and `_WD_DeleteSession`

3.
according to the ENUMS order:
```autoit
Global Enum _
		$_WD_DEBUG_None = 0, _ ; No logging to console
		$_WD_DEBUG_Error, _    ; logging to console in case of Error
		$_WD_DEBUG_Info, _     ; logging to console with additional information
		$_WD_DEBUG_Full        ; logging to console with information for developers of WebDriver UDF
```
`$_WD_DEBUG_Error` should show less information than `$_WD_DEBUG_Info` .

But currently `$_WD_DEBUG_Info` shows only URL
```autoit
	__WD_ConsoleWrite($sFuncName & ": URL=" & $sURL, $_WD_DEBUG_Info)
```
which means that `$_WD_DEBUG_Info` shows less information than `$_WD_DEBUG_Error` which shows entire `$sResponseText`
```autoit
	Local $sMessage = (($sResponseText) ? ("ResponseText=" & StringLeft($sResponseText, $_WD_RESPONSE_TRIM) & "...") : (""))
	Return SetError(__WD_Error($sFuncName, $iResult, $sMessage), 0, $sResponseText)
```

### What is the new behavior?

[...]

PR #321, opened 2022-06-01
Added $_WD_ERROR_JSON - Failed to decode request as JSON

mlipok:
### Proposed changes

capturing $_WD_ERROR_JSON - Failed to decode request as JSON

### Checklist

### Types of changes

- [ ] Other (please describe)

after such error:
`__WD_Post ==> Invalid argument (5) HTTP status = 400 : ResponseText={"value":{"error":"invalid argument","message":"Failed to decode request as JSON: ","stacktrace":"Syntax error at :1:0"}}...`

UDF set error to `$_WD_ERROR_Exception`

### What is the new behavior?

**Feature 1:**

after such error:
`__WD_Post ==> Failed to decode request as JSON (22) HTTP status = 400 : ResponseText={"value":{"error":"invalid argument","message":"Failed to decode request as JSON: ","stacktrace":"Syntax error at :1:0"}}...`

UDF set error to `$_WD_ERROR_JSON` and description to: `Failed to decode request as JSON`

Example from here: https://github.com/Danp2/au3WebDriver/pull/320#issuecomment-1143296412
```
__WD_Post ==> Failed to decode request as JSON (22) HTTP status = 400 : ResponseText={"value":{"error":"invalid argument","message":"Failed to decode request as JSON: ","stacktrace":"Syntax error at :1:0"}}...
_WD_CreateSession: {"value":{"error":"invalid argument","message":"Failed to decode request as JSON: ","stacktrace":"Syntax error at :1:0"}}
_WD_CreateSession ==> Failed to decode request as JSON (22) HTTP status = 400
```

**Feature 2:**

also this part of `__WD_DetectError` :

```autoit
Case Else
      __WD_ConsoleWrite($sFuncName & ": Not identified type of message: " & $vResult.item('message'), $_WD_DEBUG_Full)
      $iErr = $_WD_ERROR_Exception
```

helps us to catch not detected error types.

It will help in the future to create new ERROR enums which helps to minimize the need to use `$_WD_DEBUG_Full`

### Additional context

https://github.com/Danp2/au3WebDriver/pull/320#issuecomment-1143275328

https://github.com/Danp2/au3WebDriver/pull/290

### System under test

not related

### DIFF
--- a/wd_core.au3
+++ b/wd_core.au3
@@ -152,7 +152,7 @@ Global Const $aWD_ERROR_DESC[$_WD_ERROR_COUNTER] = [ _
 		"Browser or feature not supported", _
 		"Capability or value already defined", _
 		"Javascript Exception", _
-		"Failed to decode request as JSON" _
+		"JSON Exception" _
 		]
 
 Global Const $_WD_ErrorInvalidSession = "invalid session id"
--- a/wd_core.au3
+++ b/wd_core.au3
@@ -115,6 +115,7 @@ Global Enum _
 		$_WD_ERROR_NotSupported, _ ;
 		$_WD_ERROR_AlreadyDefined, _ ; Used in _WD_CapabilitiesDefine and __WD_CapabilitiesInitialize
 		$_WD_ERROR_Javascript, _ ; Javascript error
+		$_WD_ERROR_JSON, _ ; Failed to decode request as JSON
 		$_WD_ERROR_COUNTER ;
 
 Global Enum _ ; Column positions of $_WD_SupportedBrowsers
@@ -150,7 +151,8 @@ Global Const $aWD_ERROR_DESC[$_WD_ERROR_COUNTER] = [ _
 		"File issue", _
 		"Browser or feature not supported", _
 		"Capability or value already defined", _
-		"Javascript Exception" _
+		"Javascript Exception", _
+		"Failed to decode request as JSON" _
 		]
 
 Global Const $_WD_ErrorInvalidSession = "invalid session id"
@@ -233,17 +235,14 @@ Func _WD_CreateSession($sCapabilities = Default)
 
 		If @error Then
 			Local $sMessage = Json_Get($oJSON, "[value][message]")
-
-			Return SetError(__WD_Error($sFuncName, $_WD_ERROR_Exception, $sMessage), 0, "")
+			$iErr = $_WD_ERROR_JSON
 		EndIf
-	Else
-		Return SetError(__WD_Error($sFuncName, $_WD_ERROR_Exception), 0, "")
 	EndIf
 
 	; Save response details for future use
 	$_WD_SESSION_DETAILS = $sResponse
 
-	Return SetError(__WD_Error($sFuncName, $_WD_ERROR_Success), 0, $sSession)
+	Return SetError(__WD_Error($sFuncName, $iErr), 0, $sSession)
 EndFunc   ;==>_WD_CreateSession
 
 ; #FUNCTION# ====================================================================================================================
@@ -1253,7 +1252,7 @@ Func _WD_Startup()
 		EndIf
 
 		If _WinAPI_GetBinaryType($_WD_DRIVER) Then _
-			$sDriverBitness = ((@extended = $SCS_64BIT_BINARY) ? (" (64 Bit)") : (" (32 Bit)"))
+				$sDriverBitness = ((@extended = $SCS_64BIT_BINARY) ? (" (64 Bit)") : (" (32 Bit)"))
 
 		__WD_ConsoleWrite($sFuncName & ": OS:" & @TAB & @OSVersion & " " & @OSType & " " & @OSBuild & " " & @OSServicePack)
 		__WD_ConsoleWrite($sFuncName & ": AutoIt:" & @TAB & @AutoItVersion)
@@ -1662,6 +1661,8 @@ EndFunc   ;==>__WD_TranslateQuotes
 ; Example .......: No
 ; ===============================================================================================================================
 Func __WD_DetectError(ByRef $iErr, $vResult)
+	Local Const $sFuncName = "__WD_DetectError"
+
 	; Don't perform any action if error condition is
 	; already set or the webdriver result equals null
 	If $iErr Or $vResult == Null Then Return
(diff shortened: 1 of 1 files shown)


wd_core.au3
```diff
@@ -233,17 +235,14 @@ Func _WD_CreateSession($sCapabilities = Default)
 
 		If @error Then
 			Local $sMessage = Json_Get($oJSON, "[value][message]")
-
-			Return SetError(__WD_Error($sFuncName, $_WD_ERROR_Exception, $sMessage), 0, "")
+			$iErr = $_WD_ERROR_JSON
```
mlipok, line 238: should we change error form `$_WD_ERROR_Exception` to `$_WD_ERROR_JSON` after all issue related to `Json_Get($oJSON` ?

wd_core.au3
```diff
@@ -150,7 +151,8 @@ Global Const $aWD_ERROR_DESC[$_WD_ERROR_COUNTER] = [ _
 		"File issue", _
 		"Browser or feature not supported", _
 		"Capability or value already defined", _
-		"Javascript Exception" _
+		"Javascript Exception", _
+		"Failed to decode request as JSON" _
```
mlipok, line 155: maybe change this to `Failed to decode JSON` ?

wd_core.au3
Danp2: No, and I disagree with your change above. Not all errors will be JSON related. For example --

```
__WD_Post ==> Invalid argument [5] : HTTP status = 400 ResponseText={"value":{"error":"invalid argument","message":"acceptInsecureCerts is not boolean: \"xxx\"","stacktrace":""}}
_WD_CreateSession ==> Webdriver Exception [10]
```

I haven't tested it, but I was thinking about doing something like this --

```
		If @error Then
			$sMessage = Json_Get($oJSON, "[value][message]")
			__WD_DetectError($iErr, $oJSON)
```

wd_core.au3
Danp2: I would suggest something more generic, such as `JSON Exception`

Danp2:
FWIW, I'm not yet convinced that we need a JSON related error code. When you have a webdriver response like

`{"value":{"error":"invalid argument","message":"Failed to decode request as JSON: ","stacktrace":"Syntax error at :1:0"}}`

I would propose that the proper error would be `invalid argument`, which is what the existing code in __WD_DetectError would return.

mlipok:
> > What is the current behavior?
> 
> This section of your PR appears to be the new behavior rather than the current behavior. Please fix.


Thanks. Fixed.

wd_core.au3
mlipok: as so far: [revert _WD_ERROR_Exception in _WD_CreateSession()](https://github.com/Danp2/au3WebDriver/pull/321/commits/8cc11505dcd11166dac0bdd5e346290eaeaf6f5b)

mlipok:
> FWIW, I'm not yet convinced that we need a JSON related error code. When you have a webdriver response like
> 
> `{"value":{"error":"invalid argument","message":"Failed to decode request as JSON: ","stacktrace":"Syntax error at :1:0"}}`
> 
> I would propose that the proper error would be `invalid argument`, which is what the existing code in __WD_DetectError would return.


`$_WD_ERROR_InvalidArgue` is also set in many other cases, not always related to WebDriver responses.

For example:
`_WD_Cookies`
```autoit
		Case 'delete', 'deleteall'
			If $sCommand = 'delete' And IsString($sOption) = 0 Then $iErr = $_WD_ERROR_InvalidArgue
			If $sCommand = 'deleteall' And $sOption <> '' Then $iErr = $_WD_ERROR_InvalidArgue
```

`_WD_ExecuteScript`
```autoit
	Else
		$sResponse = ""
	EndIf
```

`_WD_WaitElement`
```autoit
	If $bCheckNoMatch And $iOptions <> $_WD_OPTION_NoMatch Then
	Else
```

btw.
getting response text to console is not common way to run well tested program.

In my opinion compiled exe should be delivered to non professionall user which do not want to see tons of information so `$_WD_DEBUG_Error` or at least `$_WD_DEBUG_Info` should be default for compiled program and error number/description should deliver more accurate information (I mean $_WD_ERROR_JSON instead $_WD_ERROR_Exception), without the need to looking into `sphagetii` from response text, which should be considered as an advance option only for developers, not for end user.

Other problem is that not always you can have loged all response text due to some law problems like GDPR.

As another side note, this change is in line with the change we made with `$ _WD_ERROR_Javascript` and` $ _WD_ERROR_InvalidExpression ++ CSSSelector`

mlipok:
> However, the decision was made early on with `_WD_CreateSession` to return a single error code for a failed attempt. This can be addressed by using __WD_DetectError to process the response text and extrapolate an appropriate error code.

So you are saying that the main problem in mine understanding your aproach is that `_WD_CreateSession` should always return only one kind of error, no matter what error comes back from `__WD_Post()` ?

Danp2:
> So you are saying that the main problem in mine understanding your aproach is that _WD_CreateSession should always return only one kind of error, no matter what error comes back from __WD_Post() ?

No, the main problem as I see it is that we disagree on which error code should be returned by _WD_CreateSession when this response is received --

`{"value":{"error":"invalid argument","message":"Failed to decode request as JSON: ","stacktrace":"Syntax error at :1:0"}}`

mlipok:
> which error code should be returned by _WD_CreateSession when this response is received

But for any other function `$_WD_ERROR_JSON` will be fine ?

Danp2:
> But for any other function $_WD_ERROR_JSON will be fine ?

Nope, I never said that. What I did say is [this](https://github.com/Danp2/au3WebDriver/pull/321#issuecomment-1143782363) followed by [this](https://github.com/Danp2/au3WebDriver/pull/321#issuecomment-1144235917). 

You made a case that `$_WD_ERROR_InvalidArgue` wasn't a good choice here, but failed to back it up with more details when requested.

mlipok:
What you think about this PR after this commit: [$_WD_ERROR_JSON detection moved](https://github.com/Danp2/au3WebDriver/pull/321/commits/70e5d1f21db5e5773ed6dcb98249cf4ad051a2f5)
?

Danp2:
I'm still not convinced that this new error code is needed. Can you show some examples where your latest change would apply?

mlipok:
No.
I think you have right.
I will keep in mind that in case of other JSON errors that probably appear, consider whether this new error number is needed.

PR #337, opened 2022-07-11
DemoStorage

mlipok:
### Proposed changes

added separate examples as `DemoStorage`

### Checklist

### Types of changes

- [ ] Other (please describe)

before merging new feature

### What is the new behavior?

added separate examples as `DemoStorage`

### Additional context

none

### System under test

not related

### DIFF
--- a/wd_demo.au3
+++ b/wd_demo.au3
@@ -43,6 +43,7 @@ Global $aDemoSuite[][3] = _
 		["DemoSleep", False, False], _
 		["DemoSelectOptions", False, False], _
 		["DemoStyles", False, False], _
+		["DemoStorage", False, False], _
 		["UserTesting", False, False] _
 		]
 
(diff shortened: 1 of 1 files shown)


Danp2 (commented):
This code looks familiar. 😀 A few comments --

- The `IsKeyword` coding in the ConsoleWrite is incomplete. My original plan was to use a ternary operator to display "null" when the result of the function call is null.
- Does the additional _WD_ExecuteScript at the bottom add any real value? IMO, it is performing the exact same action as _WD_Storage. How will the user known to check the browser console?
- Lastly, I question whether this example should be included in wd_demo because it is likely that most coders will never use _WD_Storage.

mlipok:
> This code looks familiar.

Yes. It is yours example.
Byt this PR I just wanted to show that we should add this as a demo.

> * The `IsKeyword` coding in the ConsoleWrite is incomplete. My original plan was to use a ternary operator to display "null" when the result of the function call is null.

If this is needed please provide fix.

> * Does the additional _WD_ExecuteScript at the bottom add any real value? IMO, it is performing the exact same action as _WD_Storage. 

yes it shows that it is accesible in browser (shows results in browser console).

> How will the user known to check the browser console?

If user want to know how to work with browser automation then he must to learn how to use browser console.

> Lastly, I question whether this example should be included in wd_demo because it is likely that most coders will never use _WD_Storage.

Good point, I was not thinking about this in such way.
Leaving it to you - you can close or merge.

Danp2:
Let's leave this out for now.

Issue #354, opened 2022-07-28
List all frames


mlipok:
### Is your feature request related to a problem? Please describe
Very often, automation concerns websites with frames.
Mostly novice users (AutoIt novice scripts) do not understand them.
Even for the more advanced ones, a diagnostic tool would be useful.

### Describe the solution you'd like

I would like to add new feature to 
_WD_FrameEnter - list all frames.
```autoit
_WD_FrameEnter($sSession, Default)
```

which should use my own JavaScript:
```javascript
FramesEnumerator()

function FramesEnumerator() {
	console.log('===PROCESSING START===');
  var result = GetFrames(window.top, '', -1, 'null', 0);
	console.log('===PROCESSING END===');
  console.log('RESULT:');
  console.log(result);
}

function GetFrames(window_current, frames_html, level_current, level_string, counter) {
  counter = counter + 1
  console.log('======== STEP = ' + counter)
  console.log('currently processing level = ' + level_current)
  console.log('currently processing "window_current":');
  console.log(window_current.location.href);

  const windows_len = window_current.length;
  const windows_max = windows_len -1
  console.log('windows_len = ' + windows_len);
  if (frames_html === '') {
	  frames_html = frames_html + level_string + '|' + window_current.location.href + '|' + '\n'
  };


  for (let i = 0; i < windows_len; i++) {
	  frames_html = frames_html + level_string + '/' + i + '|' + window_current[i].location.href + '|' + window_current.document.querySelectorAll('iframe')[i].outerHTML + '\n'
	  console.log('processing sub frames =   "' + level_string + '/' + i + '"  :');
    frames_html = GetFrames(window_current[i], frames_html, i, level_string + '/' + i, counter);
  }
  return frames_html
}
```

### Describe alternatives you've considered

Manually check the webpage

### Additional context

Example:
using for testing this website:
https://www.w3schools.com/

I get
```
null|https://www.w3schools.com/|
null/0|https://www.w3schools.com/howto/tryhow_js_slideshow_ifr.htm|<iframe src="/howto/tryhow_js_slideshow_ifr.htm" id="howto_iframe"></iframe>
null/1|about:blank|<iframe name="__tcfapiLocator" style="display: none;"></iframe>
null/2|about:blank|<iframe name="__uspapiLocator" style="display: none;"></iframe>
```
and of course it can be easily transform to AutoIt array:

<html>
<head>
	
</head>

<body>

frame position / Identifier | HREF |  element
-- | -- | --
[...]

PR #360, opened 2022-08-01
UserTesting() session validation and extra space

mlipok:
### Proposed changes

I made FR here: https://github.com/Danp2/au3WebDriver/issues/354
And I realizie that I had to add new Function() that are subject to review.
The same situation may arise if someone wants to report a problem.
That's why I found the whole region useful for testing and additional commentary.

In addition, I found that just in case it would be good to add validation if the session is correct, as basically the whole demo does not use it, and in the case of `UserTesting` scripts it would be helpful.

### Checklist

### Types of changes

- [ ] Other (please describe)

no session validation
no additionall space where user should put speciall testing function

### What is the new behavior?
added session validtation
```autoit
	_WD_CheckContext($sSession, False)
	If @error Then ; return if session is NOT OK
		ConsoleWrite("! ---> @error=" & @error & "  @extended=" & @extended & _
				" : _WD_CheckContext reported a problem with the session" & @CRLF)
		Return
	EndIf
```

additional test space inside the region:
```autoit
#Region - UserTesting
....
#EndRegion - UserTesting
```

### Additional context

none

### System under test

not related

### DIFF
--- a/wd_demo.au3
+++ b/wd_demo.au3
@@ -203,13 +203,11 @@ Func RunDemo($idDebugging, $idBrowsers, $idUpdate, $idHeadless, $idOutput)
 			ContinueLoop
 		EndIf
 
-		ConsoleWrite("+ wd_demo.au3: Running: " & $sDemoName & @CRLF)
 		_WD_CheckContext($sSession, False)
-		If @error Then ; return if session is NOT OK
-			ConsoleWrite("! ---> @error=" & @error & "  @extended=" & @extended & _
-					" : _WD_CheckContext reported a problem with the session" & @CRLF)
-			ContinueLoop
-		EndIf
+		$iError = @error
+		If @error Then ExitLoop ; return if session is NOT OK
+		
+		ConsoleWrite("+ wd_demo.au3: Running: " & $sDemoName & @CRLF)
 		If $aDemoSuite[$iIndex][2] Then
 			Call($sDemoName, $sBrowserName)
 		Else
--- a/wd_demo.au3
+++ b/wd_demo.au3
@@ -918,7 +918,15 @@ Func DemoStyles()
 
 EndFunc   ;==>DemoStyles
 
+#Region - UserTesting
 Func UserTesting() ; here you can replace the code to test your stuff before you ask on the forum
+	_WD_CheckContext($sSession, False)
+	If @error Then ; return if session is NOT OK
+		ConsoleWrite("! ---> @error=" & @error & "  @extended=" & @extended & _
+				" : _WD_CheckContext reported a problem with the session" & @CRLF)
+		Return
+	EndIf
+	
 	_WD_Navigate($sSession, 'https://www.google.com')
 	_WD_LoadWait($sSession)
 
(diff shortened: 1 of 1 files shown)


Danp2:
I can see where calling `_WD_CheckContext` could be beneficial if I were running a bunch of tests back to back, and I wanted to ensure that the session was still valid between function calls. Did you consider placing this command inside of `RunDemo` so that it will automatically execute for each test run?

mlipok:
> Did you consider placing this command inside of `RunDemo` so that it will automatically execute for each test run?

Good idea.

Danp2:
I don't see the need for the additional #region and associated commentary, but it is benign and I won't fight you on it if you insist on including it. 😆

mlipok:
This comments are not for me or you but for newebie. For this reason I would like to stay them there.

mlipok:
Please take a look on: [_WD_CheckContext - moved](https://github.com/Danp2/au3WebDriver/pull/360/commits/b3fee13337ad11c4ebb96294062025aec5e9674e)

wd_demo.au3
```diff
@@ -204,6 +204,12 @@ Func RunDemo($idDebugging, $idBrowsers, $idUpdate, $idHeadless, $idOutput)
 		EndIf
 
 		ConsoleWrite("+ wd_demo.au3: Running: " & $sDemoName & @CRLF)
+		_WD_CheckContext($sSession, False)
+		If @error Then ; return if session is NOT OK
+			ConsoleWrite("! ---> @error=" & @error & "  @extended=" & @extended & _
+					" : _WD_CheckContext reported a problem with the session" & @CRLF)
+			ContinueLoop
+		EndIf
```
Danp2, line 212: Two things wrong here IMO --

- No point in continuing the loop because rest of tests will also fail
- Should use `_RunDemo_ErrorHander` instead of custom code

Danp2 (changes requested):

wd_demo.au3
mlipok: Please take a look: [_WD_CheckContext usage](https://github.com/Danp2/au3WebDriver/pull/360/commits/42aa32cbd335de0b2be0a32120bc20d5a3ffe63b)

Danp2: I still feel there are issues here --

- No need for `$iError = @error` because you don't do anything with the stored error code
- My suggestion was to use `_RunDemo_ErrorHander`, which I believe would handle notifying the user via ConsoleWrite. Instead you simply `ExitLoop` without any type of notice to the user about why the process was aborted.

Please correct me if I'm wrong on the above items.

mlipok: > * No need for `$iError = @error` because you don't do anything with the stored error code

this code 
https://github.com/Danp2/au3WebDriver/blob/42aa32cbd335de0b2be0a32120bc20d5a3ffe63b/wd_demo.au3#L206-L208

in line 207 is setting `$iError` 
and `ExitLoop` from line 208 to line 219 
and then `$iError` is used in line 220
https://github.com/Danp2/au3WebDriver/blob/42aa32cbd335de0b2be0a32120bc20d5a3ffe63b/wd_demo.au3#L219-L222

Do my reasoning is correct ?

Danp2: Yes, you are correct.

Danp2 (approved):

Issue #254, opened 2022-03-09
wd_demo: review all examples


mlipok:
### Describe the bug

This may not be a specific bug, but we have to check all "Demos" step by step on all browsers.

### How to reproduce

Run wd_demo.au3
Set debug level to "Error"
Select all "Demos" and run them
![image](https://user-images.githubusercontent.com/11089482/157490178-edaac540-02cd-4763-a0e8-17348b2bda91.png)

### Expected behavior

Mostly there should not be any errors, but there are exceptions for example "DemoTimeouts" exptect to TimeOut error was fired.

### Screenshots

Expected TimeOut error
![image](https://user-images.githubusercontent.com/11089482/157490831-e266ce93-c346-47de-9141-f7252bd81ca1.png)

Unexpected error:
![image](https://user-images.githubusercontent.com/11089482/157490708-aa2ee6f4-a1c5-46c6-ba04-12fec191662a.png)

### Additional context

This bug report is only as a start for disscusion.

### System under test

All possible cases. If we will have time.

Danp2:
I don't get these errors when running the version of wd_demo in master branch.

mlipok:
It depends on driver, browser, windows versions.

mlipok:
And they are not all.

Danp2:
> It depends on driver, browser, windows versions.

Theoretically, it shouldn't matter. In the real world, we both know that some browsers / drivers occasionally behave differently. 

If there are certain combinations that don't work as expected, then you should list them here.

> And they are not all.

I'm not sure what you are trying to say here.

mlipok:
> If there are certain combinations that don't work as expected, then you should list them here.

Thats the plan.

mlipok:
> > If there are certain combinations that don't work as expected, then you should list them here.
> 
> Thats the plan to check and enlist them.

I see you start enlisting them here:
https://github.com/Danp2/au3WebDriver/issues/265

So I propose to close my ISSUE and sticke with your's.

Danp2:
Sorry... I forgot that this issue already existed.

mlipok:
No problem now they are connected.

mlipok:
I will check this on my way.
soon....

mlipok:
FIREFOX: **DemoElements**

```
+ wd_demo.au3: Running: DemoElements
- wd_demo.au3: #350 : Element Coords = 237 / 381.5 / 487 / 34
[...]

PR #371, opened 2022-08-08
DemoAlerts() - show that prompt was set

mlipok:
### Proposed changes

this PR is only related to the comment that describe in the specific case what kind of expected result should be to check Alert status

### Checklist

### Types of changes

- [ ] Other (please describe)

`; check status before displaying Alert`

### What is the new behavior?
`; check status before displaying Alert - this should set @error as there is no any Alert displayed yet`

### Additional context

I was checking examples for this ISSUE #254 

### System under test
not related

### DIFF
--- a/wd_demo.au3
+++ b/wd_demo.au3
@@ -508,7 +508,7 @@ EndFunc   ;==>DemoCookies
 Func DemoAlerts()
 	Local $sStatus, $sText
 
-	; check status before displaying Alert - this should set @error as there is no any Alert displayed yet
+	; check status before displaying Alert
 	$sStatus = _WD_Alert($sSession, 'status')
 	ConsoleWrite("wd_demo.au3: (" & @ScriptLineNumber & ") : " & 'Alert Detected => ' & $sStatus & @CRLF)
 
@@ -526,7 +526,7 @@ Func DemoAlerts()
 	_WD_Alert($sSession, 'Dismiss')
 
 	; show Prompt for testing
-	_WD_ExecuteScript($sSession, "prompt('User Prompt', 'Default value')")
+	_WD_ExecuteScript($sSession, "prompt('User Prompt 1', 'Default value')")
 
 	Sleep(2000)
 
@@ -534,8 +534,16 @@ Func DemoAlerts()
 	_WD_Alert($sSession, 'sendtext', 'new text')
 
 	Sleep(5000)
-	; close Alert
+	; close Alert by acceptance
 	_WD_Alert($sSession, 'Accept')
+
+	Sleep(1000)
+	; show Prompt for testing
+	_WD_ExecuteScript($sSession, "prompt('User Prompt 2', 'Default value')")
+
+	Sleep(5000)
+	; close Alert by rejection
+	_WD_Alert($sSession, 'Dismiss')
 
 EndFunc   ;==>DemoAlerts
 
--- a/wd_demo.au3
+++ b/wd_demo.au3
@@ -508,7 +508,7 @@ EndFunc   ;==>DemoCookies
 Func DemoAlerts()
 	Local $sStatus, $sText
 
-	; check status before displaying Alert
+	; check status before displaying Alert - this should set @error as there is no any Alert displayed yet
 	$sStatus = _WD_Alert($sSession, 'status')
 	ConsoleWrite("wd_demo.au3: (" & @ScriptLineNumber & ") : " & 'Alert Detected => ' & $sStatus & @CRLF)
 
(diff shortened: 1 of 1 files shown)


wd_demo.au3
```diff
@@ -508,7 +508,7 @@ EndFunc   ;==>DemoCookies
 Func DemoAlerts()
 	Local $sStatus, $sText
 
-	; check status before displaying Alert
+	; check status before displaying Alert - this should set @error as there is no any Alert displayed yet
```
Danp2, line 511: I don't see the point of this revision. @error being set is a byproduct of checking for the presence of an Alert. The main point of the `status` option is to return either True or False. Wouldn't it be better to explain that?

mlipok: In my proposed scenarion we will get:
```
+ wd_demo.au3: Running: DemoAlerts
__WD_Get ==> No alert present [12] : HTTP status = 404
wd_demo.au3: (522) : Alert Detected => False
```
instead
```
+ wd_demo.au3: Running: DemoAlerts
__WD_Get ==> No alert present [12] : HTTP status = 404
_WD_Alert ==> No alert present [12] : Parameters:   Command=status   Option=Default
wd_demo.au3: (522) : Alert Detected => False
```

Danp2: > ```autoit
> 			If $iErr = $_WD_ERROR_NoAlert Then $iErr = $_WD_ERROR_Success
> ```

Agreed... the same thought had crossed my mind

mlipok: Separate PR ? or here ?

mlipok: > The main point of the `status` option is to return either True or False. Wouldn't it be better to explain that?

If we change error handling then the comments change in this regards is not needed as there is already status reported to console
```
$sStatus = _WD_Alert($sSession, 'status')
ConsoleWrite("wd_demo.au3: (" & @ScriptLineNumber & ") : " & 'Alert Detected => ' & $sStatus & @CRLF)
```

mlipok:
I have further fefactoring for this DemoAlerts()...

wd_demo.au3
mlipok: > Separate PR ? or here ?

I create separate PR later

mlipok:
As for me it is all in this PR, please review again and merge if you agree with my opinion about the comments:
> > The main point of the `status` option is to return either True or False. Wouldn't it be better to explain that?
> 
> If we change error handling then the comments change in this regards is not needed as there is already status reported to console
> 
> ```autoit
> $sStatus = _WD_Alert($sSession, 'status')
> ConsoleWrite("wd_demo.au3: (" & @ScriptLineNumber & ") : " & 'Alert Detected => ' & $sStatus & @CRLF)
> ```

wd_demo.au3
Danp2: > Separate PR ? or here ?

Here is fine with me.

> If we change error handling then the comments change in this regards is not needed as there is already status reported to console

True

Danp2:
> using Storage to show that prompt was set 

Why is this needed here? I don't see the purpose of using "storage" in the middle of an Alerts demo.

Danp2:
Sorry, but I don't understand how this code demonstrates that. With this line --

` 	_WD_ExecuteScript($sSession, "window.localStorage.setItem('testing',prompt('User Prompt 1', 'Default value'))") `

you introduce advanced concepts that have nothing to do with _WD_Alert IMO.

mlipok:
Yes it is more advanced than only showing prompt.
Because in order to check if ti was properly set by prompt it stores the value in webbrowser and then we check the stored value using  `_WD_Storage($sSession, 'testing')`

mlipok:
https://github.com/Danp2/au3WebDriver/blob/a41396be2d1a71cd3d1015cafbb01c164124726e/wd_demo.au3#L531-L532
popup prompt to input data, which is saved in browser `Storage` 

https://github.com/Danp2/au3WebDriver/blob/a41396be2d1a71cd3d1015cafbb01c164124726e/wd_demo.au3#L542-L544
check what trully was received and save in browser Storage

I call this `validation`.

Danp2:
> I call this validation.

AFAICS you aren't validating the functionality of _WD_Alert. Therefore, this doesn't belong in DemoAlerts function.

>  By the way, it also shows how Storage can be used.

Also doesn't belong here.

mlipok:
> AFAICS you aren't validating the functionality of _WD_Alert. Therefore, this doesn't belong in DemoAlerts function.

So how we can validate that 

```js
var userinput = prompt('User Prompt 1', 'Default value')
```

browser get the value provided by entering string to the prompt ?

Danp2:
You need a webpage that accepts the user input and then displays it on screen like they do [here](https://www.quanzhanketang.com/jsref/tryjsref_prompt.html?filename=tryjsref_prompt). I haven't tried this yet, but I was thinking that you could open a new tab and then use _WD_ExecuteScript to update the tab's source code to mimic the behavior of that website without the frames.

Danp2 (approved):

Issue #368, opened 2022-08-06
_WD_ElementSelectAction should validate Multiselect request


Danp2:
### Describe the bug

There is currently no validation taking place when executing the Multiselect option of _WD_ElementSelectAction 

### How to reproduce

Steps to reproduce the behavior (e.g.):

```autoit
Local $a_Printer[] = ['Save as PDF']
_WD_ElementSelectAction($WD_SESSION, $sSelectElement, "MULTISELECT", $a_Printer)
```

See link below for full context 

### Expected behavior

Function should verify that that calling routine supplied multiple options to match. Also, the Select element should be queried for `multiselect` capability and return an error if it isn't.

### Additional context

https://www.autoitscript.com/forum/topic/205553-webdriver-udf-help-support-iii/?do=findComment&comment=1505272

### System under test

Please complete the following information.

- OS: [e.g. Windows 10]
- OS Arch.: [e.g. X64]
- Browser [e.g. firefox]
- Browser version [e.g. 96.0.3]

mlipok:
I plan to handle this issue in https://github.com/Danp2/au3WebDriver/pull/367

mlipok:
> Also, the Select element should be queried for `multiselect` capability and return an error if it isn't.

In case of error should any selecting/dispatch action should be taken on `<select>` element ?
What kind of error you propose ?

Danp2:
> In case of error should any selecting/dispatch action should be taken on select element?

No. This validation should take place before performing any other actions.

>  What kind of error you propose ?

I haven't thought about it TBH. Perhaps `$_WD_ERROR_InvalidArgue` or `$_WD_ERROR_ElementIssue`.

mlipok:
`$_WD_ERROR_ElementIssue` sounds better.

mlipok:
https://www.autoitscript.com/forum/topic/205553-webdriver-udf-help-support-iii/?do=findComment&comment=1505272
```
this discussion has pointed out a few weak points with the current "multiselect" implementation --

No validation that the target select element supports multiselect
No validation that multiple selections were requested
```

It is suplemented here:
https://github.com/Danp2/au3WebDriver/pull/367/commits/bf1c64807eab2130e27f685a2d59be6c49e6fd9f

https://github.com/Danp2/au3WebDriver/blob/bf1c64807eab2130e27f685a2d59be6c49e6fd9f/wd_helper.au3#L1030-L1032

Danp2:
[...]

PR #367, opened 2022-08-05
_WD_ElementSelectAction + SINGLESELECT

mlipok:
### Proposed changes
This PR add feature to easily select single `<option>` element by its label

### Checklist

Put an `x` in the boxes that apply. If you're unsure about any of them, don't hesitate to ask. We are here to help!<br>
This is simply a reminder of what we are going to look for before merging your code.

- [x] I have read and noticed the [CODE OF CONDUCT](https://github.com/Danp2/au3WebDriver/blob/master/docs/CODE_OF_CONDUCT.md) document
- [x] I have read and noticed the [CONTRIBUTING](https://github.com/Danp2/au3WebDriver/blob/master/docs/CONTRIBUTING.md) document
- [ ] I have added necessary documentation or screenshots (if appropriate)

### Types of changes

Please check `x` the type of change your PR introduces:

- [ ] Bugfix (change which fixes an issue)
- [x] Feature (change which adds functionality)
- [ ] Code style update (formatting, renaming)
- [x] Refactoring (functional, structural)
- [ ] Documentation content changes
- [ ] Other (please describe)

### What is the current behavior?

no way for easy way to select single option by LabelName

### What is the new behavior?
1.

can be easily done with
```autoit
_WD_ElementSelectAction($sSesion, $sSelectElement, 'singleLabel', 'LabelNameToSeclect')
```

2.
Additionally `_WD_ExecuteScript` currently returns true or false as JavaScript was refactored

3.
JavaScript code was refactored to use `functions` because this is easier to copy them to browser and test and maintain in the future.

4.
JavaScript code is stored as static value which optimize speed of AutoIt execution in case of using this function multiple times.

5.
[Function should verify that that calling routine supplied multiple options to match.](https://github.com/Danp2/au3WebDriver/issues/368#issuecomment-1207175954)

### Additional context
https://www.autoitscript.com/forum/topic/205553-webdriver-udf-help-support-iii/?do=findComment&comment=1505272

https://github.com/Danp2/au3WebDriver/issues/368#issuecomment-1207175954

### System under test

not related

PR #373, opened 2022-08-10
_WD_ElementSelectAction 'selectAll' JS validation

mlipok:
### Proposed changes
Functions should have more validation / error checking.
In case when there is no need to select options because they are already selected there should not be done any change to `HTML DOM` so also event should not be fired.

### Checklist

### Types of changes

- [ ] Other (please describe)

JS code just going on without any validation

### What is the new behavior?

1.
check if `SelectElement` has `multiple` atribute and set error to `$_WD_ERROR_ElementIssue` if there element is not `multiple` able.

2.
return false when no selection was taken, and not fires change event in such case.

### Additional context
https://github.com/Danp2/au3WebDriver/issues/368#issuecomment-1209081023

### System under test
not related

### DIFF
--- a/wd_helper.au3
+++ b/wd_helper.au3
@@ -1058,20 +1058,20 @@ Func _WD_ElementSelectAction($sSession, $sSelectElement, $sCommand, $aParameters
 							"		return '';" & _
 							"	}; " & _
 							"	let options = SelectElement.options;" & _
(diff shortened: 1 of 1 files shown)


mlipok:
already checked with DemoSelectOptions() from this version: https://github.com/Danp2/au3WebDriver/pull/367/files

PR #376, opened 2022-08-10
_WD_ElementSelectAction 'selectedLabels' JS refact

mlipok:
### Proposed changes

refactoring JS for better code maintaince.

### Checklist

### Types of changes

- [ ] Other (please describe)

hard to move and test in browser Console

### What is the new behavior?

easier JS code maintaince

### Additional context

none

### System under test
not related

### DIFF
--- a/wd_helper.au3
+++ b/wd_helper.au3
@@ -1067,12 +1067,19 @@ Func _WD_ElementSelectAction($sSession, $sSelectElement, $sCommand, $aParameters
 					$iErr = @error
 
 				Case 'selectedLabels'
-					$sScript = _
-							"var result ='';" & _
-							"var options = arguments[0].selectedOptions;" & _
-							"for (let i = 0; i < options.length; i++)" & _
-							" {result += options[i].label + '\n'};" & _
-							"return result;"
+					$sScript = StringReplace( _
+							"function GetSelecteLabels(SelectElement) {" & _
+							"	let result ='';" & _
+							"	let options = SelectElement.selectedOptions;" & _
+							"	for (let i = 0, o; i < options.length; i++)	{" & _
+							"		o = options[i];" & _
+							"		result += o.label + '\n';" & _
+							"	};" & _
+							"	return result;" & _
+							"};" & _
+							"var SelectElement = arguments[0];" & _
+							"return GetSelecteLabels(SelectElement);" & _
+							"", @TAB, '')
 					$vResult = _WD_ExecuteScript($sSession, $sScript, __WD_JsonElement($sSelectElement), Default, $_WD_JSON_Value)
 					$iErr = @error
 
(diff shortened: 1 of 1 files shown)


Danp2:
> refactoring JS for better code maintaince.

Some questions / comments about the refactored code --

- You switched from `var` to `let` when declaring variables. Is there some technical reason to do this that makes it better or more efficient? I'm not aware of any.

- The code "expanded" from 5 lines to 11 lines of JS. How does this make it easier to maintain?

- You should rename the script variable and declare it with `local static` so that the StringReplace only occurs once.

mlipok:
> * You switched from `var` to `let` when declaring variables. Is there some technical reason to do this that makes it better or more efficient? I'm not aware of any.

Recently I was studing JS thus I change some of my habbits.
using analogy to AutoIt   `let` is local scope and `var` is global scope


> * The code "expanded" from 5 lines to 11 lines of JS. How does this make it easier to maintain?

["Program are meant to be read by humans and only incidentally for computers and execute" - Donald Knuth, "The Art of Computer Programming"](https://youtu.be/c-kav7Tf834)

recently 'multiSelect' was few time refactored, each time I had to copy the JS to browser console and maintain JS code.
For this reason 'multiSelect' was refactored to use functions and expose SelectElement out of the function.

How it relates to other one like 'selectedLabels' ? I wanted to refactor all other functions to have use the same coding styles and the same variable names.



> * You should rename the script variable and declare it with `local static` so that the StringReplace only occurs once.

Should I also move all other JScirpt templates at top like `$sScript_MultiSelectTemplate` was moved ?

mlipok:
already checked with DemoSelectOptions() from this version: https://github.com/Danp2/au3WebDriver/pull/367/files

Danp2:
>> You switched from var to let when declaring variables. Is there some technical reason to do this that makes it better or more efficient? I'm not aware of any.

> Recently I was studing JS thus I change some of my habbits.using analogy to AutoIt let is local scope and var is global scope

AFAICS, your change had no effect because the variables are scoped at the function level, which is exactly what `var` does within a function. 

> Should I also move all other JScirpt templates at top like $sScript_MultiSelectTemplate was moved ?

I don't think they need to be moved.

mlipok:
> AFAICS, your change had no effect because the variables are scoped at the function level, which is exactly what `var` does within a function.

Technically you are right
however, IMO: taking into account the correctness of coding in JS (something like good coding practice), it should be changed .



> I don't think they need to be moved.

ok

Danp2 (changes requested):
You should rename the script variable and declare it with local static so that the StringReplace only occurs once.

mlipok:
> You should rename the script variable and declare it with local static so that the StringReplace only occurs once.
Done: [Static $sScript_SelectedLabelsTemplate](https://github.com/Danp2/au3WebDriver/pull/376/commits/8bbfe4318fd9aa2fc7852b9534b1ab3137c7fc09)

Danp2 (approved):

PR #379, opened 2022-08-17
$_WD_SupportedBrowsers should be const

mlipok:
### Proposed changes

Code CleanUp and Good Coding practice

### Checklist

### Types of changes

- [x] Code style update (formatting, renaming)
- [ ] Other (please describe)

`$_WD_SupportedBrowsers` is declared as `Global`

### What is the new behavior?
`$_WD_SupportedBrowsers` is declared as `Global Const`

`$_WD_BROWSER` was moved as they are related to `$_WD_SupportedBrowsers`

`$_WD_ERROR_COUNTER` is renamed to `$_WD_ERROR__COUNTER` I mean there was added underscore to have better `InteliCode` support. I mean  `$_WD_ERROR_COUNTER`  is not normal variable and should not be used by user thus, typing ``$_WD_ERROR_C` should not popup this variable by `InteliCode` engine, especially when especially for the new upcoming version of SciTE4AutoIt3.

### Additional context

none

### System under test

not related

### DIFF
--- a/wd_core.au3
+++ b/wd_core.au3
@@ -117,17 +117,6 @@ Global Enum _
 		$_WD_ERROR_Javascript, _ ; Javascript error
 		$_WD_ERROR_COUNTER ; Defines row count for $aWD_ERROR_DESC
 
-Global Enum _ ; Column positions of $_WD_SupportedBrowsers
-		$_WD_BROWSER_Name, _
-		$_WD_BROWSER_ExeName, _
-		$_WD_BROWSER_DriverName, _
-		$_WD_BROWSER_64Bit, _
-		$_WD_BROWSER_OptionsKey, _
-		$_WD_BROWSER_LatestReleaseURL, _
-		$_WD_BROWSER_LatestReleaseRegex, _
-		$_WD_BROWSER_NewDriverURL, _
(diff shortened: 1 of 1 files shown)


Danp2:
> $_WD_SupportedBrowsers is declared as Global Const

Fine

> $_WD_BROWSER was moved as they are related to $_WD_SupportedBrowsers

Ok

> $_WD_ERROR_COUNTER is renamed to $_WD_ERROR__COUNTER 

Wouldn't the "standard" here be to put the double underscore at the front instead of in the middle (ie: $__WDVERSION)?

mlipok:
> Wouldn't the "standard" here be to put the double underscore at the front instead of in the middle (ie: $__WDVERSION)?

I think no.
Because the prefix `$_WD_ERROR_*******` should be the same.
This is kind of reminder something like  **Hey this `$_WD_ERROR__COUNTER`** be used as they are used only internally by the UDF.

So using `InteliSens` 

to use them you should press underscore twice
image

Using single underscore it also be listed but not in the middle
image

but at the end
image

Leaving them like I proposed it is still visible in the same prefixed `$_WD_ERROR_` group of variables, and as this is on the end this is good ..... as for me ;)

Issue #388, opened 2022-08-24
_WD_FrameList / Frame visibility


Danp2:
As TODO in the future:

_WD_FrameList should return information about visibility of the frame

for example:
https://stackoverflow.com/questions/19669786/check-if-element-is-visible-in-dom

![image](https://user-images.githubusercontent.com/11089482/186519722-2d7d707a-e1d4-49a4-af2b-92a05e6c4595.png)

@Danp2 if you agree plese reference this as new issue, as I think that, for now we should focus on merging this what we have currently already developed.

_Originally posted by @mlipok in https://github.com/Danp2/au3WebDriver/issues/362#issuecomment-1226302749_

mlipok:
you can assign this issue to me

mlipok:
Should we also get informationa about shadow roots ?

Danp2:
Definitely not as part of _WD_FrameList. Please open a separate issue if you would like to discuss further.

Issue #27, opened 2019-11-01
Review / enhance functionality of _WD_Frame* functionality


Danp2:
- Proper setting of @error flag
- Standardization with other UDF functions

mlipok:
Is tis still valid issue ?

Danp2:
_WD_FrameEnter should be fine after all the recent updates. Looking at _WD_FrameLeave, I think it would be wise to review the coding associated with processing the webdriver response.

Danp2:
AFAICS, all browsers are returning the same result for a successful switch to parent frame.

#### Firefox
```
__WD_Post: URL=HTTP://127.0.0.1:4444/session/78ef96d8-6bbd-472a-a1da-f43a5880fc29/frame/parent; Data={}
__WD_Post ==> Success [0] : HTTP status = 200 ResponseText={"value":null}
_WD_Window ==> Success [0] : Parameters:   Command=parent   Option={}
_WD_FrameLeave ==> Success [0]
```
#### Chrome
```
__WD_Post: URL=HTTP://127.0.0.1:9515/session/df4215a669b23dbedcf1da8927935b6a/frame/parent; Data={}
__WD_Post ==> Success [0] : HTTP status = 200 ResponseText={"value":null}
_WD_Window ==> Success [0] : Parameters:   Command=parent   Option={}
_WD_FrameLeave ==> Success [0]
```
#### Edge
```
__WD_Post: URL=HTTP://127.0.0.1:9515/session/eb4e5d6887c1c4bd9f2c76912f53f41a/frame/parent; Data={}
__WD_Post ==> Success [0] : HTTP status = 200 ResponseText={"value":null}
_WD_Window ==> Success [0] : Parameters:   Command=parent   Option={}
_WD_FrameLeave ==> Success [0]
```
#### Opera
```
__WD_Post: URL=HTTP://127.0.0.1:9515/session/266228571a93c951178f846ac1451c51/frame/parent; Data={}
__WD_Post ==> Success [0] : HTTP status = 200 ResponseText={"value":null}
_WD_Window ==> Success [0] : Parameters:   Command=parent   Option={}
_WD_FrameLeave ==> Success [0]
```

Issue #20, opened 2018-06-16
_WD_CreateSession fails to return session ID with Microsoft web driver


Danp2:
Originally reported here -- https://www.autoitscript.com/forum/topic/191990-webdriver-udf-w3c-compliant-version-05132018/?do=findComment&comment=1394743

Danp2:
Found https://github.com/w3c/webdriver/issues/1201, which indicates that that correct implementation is for the sessionid to be within the `value` key

Danp2:
Confirmed working in v10.0.17763.1 of microsoftwebdriver.exe

Issue #19, opened 2018-05-24
SetTimeout unknown error: 'ms' must be a double


Danp2:
From [here](https://www.autoitscript.com/forum/topic/191990-webdriver-udf-w3c-compliant-version-05132018/?do=findComment&comment=1392309), 

> I tried the example code with chrome and received error when testing timeouts 'SetTimeout unknown error: 'ms' must be a double'.

Danp2:
Per new information reported [here](https://www.autoitscript.com/forum/topic/192730-webdriver-udf-help-support/?page=5&tab=comments#comment-1394251), the following works with chrome --

`_WD_Timeouts($sSession, '{"type":"page load","ms":2000}')`

Danp2:
Confirmed fixed in Chromedriver v2.42

PR #393, opened 2022-08-29
_WD_FrameList - added Frame visibility info

mlipok:
### Proposed changes

_WD_FrameList should return information about visibility of the frame

for example:
https://stackoverflow.com/questions/19669786/check-if-element-is-visible-in-dom

image

### Checklist

### Types of changes

- [ ] Other (please describe)

No information about FrameVisibility

### What is the new behavior?

added `_WD_FRAMELIST_FrameVisibility` and new column `IsHidden` to the resulted array 

image

### Additional context

was discussed here: https://github.com/Danp2/au3WebDriver/issues/388

### System under test

not relatead


wd_helper.au3
```diff
@@ -857,16 +859,26 @@ Func __WD_FrameList_Internal($sSession, $sLevel, $sFrameAttributes, $_WD_DEBUG_C
...
-					$sMessage = 'Error occured on "' & $sLevel & '" level when trying to check atributes child frames #' & $iFrame
+					$sMessage = 'Error occured on "' & $sLevel & '" level when trying to check attributes child frames #' & $iFrame
 				Else
-					$sFrameAttributes = StringRegExpReplace($sFrameAttributes, '\R', '')
-					$vResult &= __WD_FrameList_Internal($sSession, $sLevel & '/' & $iFrame, $sFrameAttributes, $_WD_DEBUG_Calling)
+					Local $sScript = _
+							"var element = document.querySelectorAll('iframe')[" & $iFrame & "];" & _
+							"var style = window.getComputedStyle(element);" & _
+							"return (style.display === 'none');"
+					$bIsHidden = _WD_ExecuteScript($sSession, $sScript, Default, Default, $_WD_JSON_Value)
```
Danp2, line 868: Have you tested this? AFAICS, this won't be accurate when a parent container is hidden and not the actual frame.

mlipok: If parent container is hidden, do you expect that actual fram will be visible?

mlipok: Take a look on 
https://user-images.githubusercontent.com/11089482/186519722-2d7d707a-e1d4-49a4-af2b-92a05e6c4595.png

Is this the exact case you are refering to ?

Danp2: Yes, that is one instance. But you can't just check the parent element if that's what you are thinking.

mlipok: > Take a look on https://user-images.githubusercontent.com/11089482/186519722-2d7d707a-e1d4-49a4-af2b-92a05e6c4595.png
> 
> Is this the exact case you are refering to ?

IMHO this case should show 'IsHidden' status as true.

mlipok: > But you can't just check the parent element if that's what you are thinking.

No this is not my idea.

Danp2: > IMHO this case should show 'IsHidden' status as true.

Correct... I didn't mean to imply otherwise

> No this is not my idea.

ok

mlipok: So I do not understand.

mlipok: > this won't be accurate when a parent container is hidden and not the actual frame.

What you are refering here ?

Danp2: >>   this won't be accurate when a parent container is hidden and not the actual frame.

> What you are refering here ?

In this case, `a parent container` refers to the parent element *and* any of its ancestral parent elements. This is why I said "But you can't just check the parent element if that's what you are thinking" because simply checking the parentElement's visibility isn't sufficient to ensure that the frame is visible.

Danp2: You may want to test the `Displayed` functionality of _WD_ElementAction to see if it works correctly for this.

mlipok: Do you mean that we should check follow this: https://www.w3.org/TR/webdriver/#element-displayedness
`The visibility of an [element](https://www.w3.org/TR/webdriver/#dfn-elements) is guided by what is perceptually visible to the human eye. In this context, an [element](https://www.w3.org/TR/webdriver/#dfn-elements)’s displayedness does not relate to the [visibility](https://www.w3.org/TR/webdriver/#dfn-visibility) or [display](https://www.w3.org/TR/webdriver/#dfn-display) style properties.`

and use `_WD_ElementAction` with `DISPLAYED` instead checking it via `HTML DOM` ?

mlipok: Does this mean that element that is visible in `HTML DOM` but is not visible to the human eyes because it is on the 3rd page and to make it visible current view should be scroled down.... is treated differently?

differently... when I use 

- `HTML DOM` which should say is not hidden
- WEBDRIVER which should say is not visible to human eye, as this is far, far away on the bottom of the page.

?

mlipok:
Working on `_WD_ElementAction` with `DISPLAYED`

As for now I have:
image

mlipok:
I get into a issue with getting value:
```
__WD_Post: URL=HTTP://127.0.0.1:4444/session/422c9a63-392f-4651-b0ac-ec2069ab9a45/execute/sync; Data={"script":"return document.querySelectorAll('iframe')[0];", "args":[]}
__WD_Post ==> Success [0] : HTTP status = 200 ResponseText={"value":{"element-6066-11e4-a52e-4f735466cecf":"57c69f02-ae5c-4234-992e-09525f059044"}}
_WD_ExecuteScript ==> Success [0]
```
you can test it with:
```autoit
Local $sResponse = '{"value":{"element-6066-11e4-a52e-4f735466cecf":"57c69f02-ae5c-4234-992e-09525f059044"}}'
Local $oJSON = Json_Decode($sResponse)
ConsoleWrite("! " & @CRLF)

Local $sValue = Json_Get($oJSON, '[value]')
ConsoleWrite("> "  & $sValue & @CRLF)
```
my result is:
```
+>Setting Hotkeys...--> Press Ctrl+Alt+F5 to Restart. --> Press Ctrl+Shift+F5 to Stop.
! 
> 
+>14:35:11 AutoIt3 ended. rc:0
```

Why I can't get value ?

Danp2:
> Why I can't get value ?

Try this instead --

```autoit
#include "wd_core.au3"

Local $sResponse = '{"value":{"element-6066-11e4-a52e-4f735466cecf":"57c69f02-ae5c-4234-992e-09525f059044"}}'
Local $oJSON = Json_Decode($sResponse)
ConsoleWrite("! " & @CRLF)

Local $sValue = Json_Get($oJSON, $_WD_JSON_Element)
ConsoleWrite("> "  & $sValue & @CRLF)
```

mlipok:
Thanks.
That do this trick.
image

mlipok:
My tests show that there are differences
image

Danp2:
> My tests show that there are differences

Going back to my earlier statement --

> In this case, a parent container refers to the parent element and any of its ancestral parent elements. This is why I said "But you can't just check the parent element if that's what you are thinking" because simply checking the parentElement's visibility isn't sufficient to ensure that the frame is visible.

Have you verified which one is providing the correct result?

wd_helper.au3
mlipok: > > > this won't be accurate when a parent container is hidden and not the actual frame.
> 
> > What you are refering here ?
> 
> In this case, `a parent container` refers to the parent element _and_ any of its ancestral parent elements. This is why I said "But you can't just check the parent element if that's what you are thinking" because simply checking the parentElement's visibility isn't sufficient to ensure that the frame is visible.

On the follwing screenshot frame `#19  null/13/0` is hidden thus all sub frames `null/13/0/.......`  in this case `#20...#27` are also hidden and only `WebDriver test` shows  IsHiddenWebDriver=True but `JS test` shows IsHiddenJS=False thus ... `IsHiddenJS` shows invalid values.

This is what you was thinking about ?

image

mlipok: If so ... What I should do ? to leave only `IsHiddenWebDriver` feature and remove `IsHiddenJS` feature ?

mlipok:
> > My tests show that there are differences
> 
> Going back to my earlier statement --
> 
> > In this case, a parent container refers to the parent element and any of its ancestral parent elements. This is why I said "But you can't just check the parent element if that's what you are thinking" because simply checking the parentElement's visibility isn't sufficient to ensure that the frame is visible.
> 
> Have you verified which one is providing the correct result?

take a look here: https://github.com/Danp2/au3WebDriver/pull/393#discussion_r1119840031

wd_helper.au3
Danp2: I suggest removing `IsHiddenJS` if it isn't accurate.

mlipok: Take a look again: [hidden status only from WebDriver - not JS](https://github.com/Danp2/au3WebDriver/pull/393/commits/bf60fce8e067e16a25b01ab6af054fe9ef433ce8)

mlipok:
btw.
I think that we can close realted ISSUE: https://github.com/Danp2/au3WebDriver/issues/388

Danp2:
@mlipok I added a link so that the issue will close once the PR is merged.

wd_helper.au3
mlipok: Can we cosider this part as resolved ?

wd_demo.au3
```diff
@@ -662,33 +664,38 @@ Func DemoFrames()
...
+	_ArrayDisplay($aFrameList, 'Example 6v2 - stackoverflow.com - check if it is still relative to "null/2"', 0, 0, Default, $sArrayHeader)
 
 	; check if document context location is Top Window
 	ConsoleWrite("> " & @ScriptLineNumber & " IsWindowTop = " & _WD_IsWindowTop($sSession) & @CRLF)
 
+	; now lets try to check location of multiple elements on multiple frames
+	; go to website
+	_WD_Navigate($sSession, 'https://www.tutorialspoint.com/html/html_frames.htm#')
+	_WD_LoadWait($sSession)
+
```
Danp2, line 698: This appears to be unfinished.

Danp2 (changes requested):

wd_demo.au3
mlipok: working on

mlipok:
image

`#Region - Testing element location in frame set`
is related to planed `_WD_LocateElement()` function, which will be in separate PR soon.

wd_demo.au3
Danp2: This still doesn't appear complete. Is Ex7 related to `_WD_LocateElement()`? If so, then why not submit it as part of that future PR instead of this one?

mlipok: Checked. You have right.
Please take a look on: [removed Example 7](https://github.com/Danp2/au3WebDriver/pull/393/commits/9ed605ccc53fceb9d731b7aa482440a5041f104f)

Danp2 (approved):

mlipok:
You can merge, which will let me do next PR with `_WD_LocateElement()`

PR #396, opened 2022-08-31
_WD_LinkClickByText: Add starting element

Danp2:
### Proposed changes

Enhance the existing _WD_LinkClickByText function by optionally allowing the specification of a starting element.

### Checklist

### Types of changes

- [ ] Other (please describe)

Entire webpage is searched

### What is the new behavior?

Uses optional Starting Element parameter of _WD_FindElement to allow the search to be limited to a specific section of the webpage

### System under test

- OS: [e.g. Windows 10]
- OS Arch.: [e.g. X64]
- Browser [e.g. firefox]

### DIFF
--- a/wd_helper.au3
+++ b/wd_helper.au3
@@ -260,10 +260,11 @@ EndFunc   ;==>_WD_Attach
 ; #FUNCTION# ====================================================================================================================
 ; Name ..........: _WD_LinkClickByText
 ; Description ...: Simulate a mouse click on a link with text matching the provided string.
-; Syntax ........: _WD_LinkClickByText($sSession, $sText[, $bPartial = Default])
-; Parameters ....: $sSession - Session ID from _WD_CreateSession
-;                  $sText    - Text to find in link
-;                  $bPartial - [optional] Search by partial text? Default is True
+; Syntax ........: _WD_LinkClickByText($sSession,  $sText[,  $bPartial = Default[,  $sStartElement = Default]])
+; Parameters ....: $sSession      - Session ID from _WD_CreateSession
+;                  $sText         - Text to find in link
+;                  $bPartial      - [optional] Search by partial text? Default is True
+;                  $sStartElement - [optional] Element ID of element to use as starting point
 ; Return values .: Success - None.
 ;                  Failure - "" (empty string) and sets @error to one of the following values:
 ;                  - $_WD_ERROR_Exception
@@ -277,10 +278,11 @@ EndFunc   ;==>_WD_Attach
 ; ===============================================================================================================================
 Func _WD_LinkClickByText($sSession, $sText, $bPartial = Default, $sStartElement = Default)
 	Local Const $sFuncName = "_WD_LinkClickByText"
-	Local Const $sParameters = 'Parameters:   Text=' & $sText & '   Partial=' & $bPartial
+	Local Const $sParameters = 'Parameters:   Text=' & $sText & '   Partial=' & $bPartial & '   StartElement=' & $sStartElement
 	Local $bIsVisible = False, $bIsEnabled = False
 
 	If $bPartial = Default Then $bPartial = True
+	If $sStartElement = Default Then $sStartElement = ""
 
 	Local $aElements = _WD_FindElement($sSession, ($bPartial) ? $_WD_LOCATOR_ByPartialLinkText : $_WD_LOCATOR_ByLinkText, $sText, $sStartElement, True)
 	Local $iErr = @error
--- a/wd_helper.au3
+++ b/wd_helper.au3
@@ -284,27 +284,15 @@ Func _WD_LinkClickByText($sSession, $sText, $bPartial = Default, $sStartElement 
 	If $bPartial = Default Then $bPartial = True
 	If $sStartElement = Default Then $sStartElement = ""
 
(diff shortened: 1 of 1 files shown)


wd_helper.au3
```diff
@@ -260,10 +260,11 @@ EndFunc   ;==>_WD_Attach
...
 ; Description ...: Simulate a mouse click on a link with text matching the provided string.
-; Syntax ........: _WD_LinkClickByText($sSession, $sText[, $bPartial = Default])
-; Parameters ....: $sSession - Session ID from _WD_CreateSession
-;                  $sText    - Text to find in link
-;                  $bPartial - [optional] Search by partial text? Default is True
+; Syntax ........: _WD_LinkClickByText($sSession,  $sText[,  $bPartial = Default[,  $sStartElement = Default]])
+; Parameters ....: $sSession      - Session ID from _WD_CreateSession
+;                  $sText         - Text to find in link
+;                  $bPartial      - [optional] Search by partial text? Default is True
+;                  $sStartElement - [optional] Element ID of element to use as starting point
```
mlipok, line 267: Element ID of element (from _WD_FindElement or _WD_WaitElement)

mlipok: other than that one remark, everything else seems okay.

Danp2: I disagree. IMO, it shouldn't be necessary to list these functions by name. Plus there are other functions that return an Element ID, ie: _WD_GetMouseElement, _WD_GetElementFromPoint, _WD_GetShadowRoot, etc.

mlipok: I will review this tomorrow.

mlipok: Ok.
Agree bo all other `Element ID of element` wording should be reviewed, as for example: `_WD_GetShadowRoot` is using such description: `$sStartElement - [optional] a string value. Default is ""`

If you agree please create related issue.

btw.
this PR #396 is fine, as for me.

Danp2: AFAICS, `_WD_GetShadowRoot` is the only function where this needs to be revised.

mlipok: > AFAICS, `_WD_GetShadowRoot` is the only function where this needs to be revised.

also description for return vaule in `_WD_GetElementByRegEx`

mlipok: and maybe `$sStartNodeID - [optional] Element ID to use as starting node. Default is ""` from `_WD_FindElement`

mlipok: also there are few places where description like `Element ID from _WD_FindElement`  is already used.

Danp2: What are you thoughts about changing all instances of `$sStartElement` to `$sStartNodeID` so that the parameter naming is "standardized"?

mlipok: Good idea.
Please take care of this.

mlipok: the same with `$sElement` >> `$sElementNodeID` ?

mlipok: for example 

in `_WD_LoadWait`
current:
`;                  $sElement - [optional] Element ID (from _WD_FindElement or _WD_WaitElement) to confirm DOM invalidation`
proposal:
`;                  $sElementNodeID - [optional] ElementNodeID to confirm DOM invalidation`


in `_WD_FindElement`
current:
`; Return values .: Success - ElementNodeID(s) returned by web driver.`
proposal:
`; Return values .: Success - ElementNodeID(s) returned by web driver.`

Danp2: > the same with $sElement >> $sElementNodeID ?

Not sure about that. This is the description for $sStartNodeID --

`;                  $sStartNodeID - [optional] Element ID to use as starting node. Default is ""`

This make sense to me because the parameter name matches the description (although the word `element` isn't utilitzed).

I don't like the use of `element` and `node` together in the same variable name because they are essentially the same thing IMO. One could argue that `$sElement` >> `$sElementID` would be more consistent naming.

mlipok: > One could argue that `$sElement` >> `$sElementID` would be more consistent naming.

This would be good.

mlipok: But what about such change:

in _WD_FindElement
current:
; Return values .: Success - Element ID(s) returned by web driver.
proposal:
; Return values .: Success - ElementID(s) returned by web driver.

and in `_WD_ElementOptionSelect`
current:
`;                  $sStartElement - [optional] Element ID of element to use as starting point`
proposal:
`;                  $sStartNodeID - [optional] ElementID of element to use as starting point`
or better proposal:
`;                  $sStartElementID - [optional] ElementID of element to use as starting point`

it would stadarize to 
`ElementID` and `$sStartElementID` and `$sElementID`

without spaces `Element ID`

mlipok: btw.
https://stackoverflow.com/a/9979779/5314940
`An element is a specific type of node,`

can `_WD_FIndElement` finds any node types ?
or only `Node.ELEMENT_NODE (1)` https://developer.mozilla.org/en-US/docs/Web/API/Node/nodeType

mlipok: > I don't like the use of `element` and `node` together in the same variable name because they are essentially the same thing IMO.

In this reference documentation:
https://developer.mozilla.org/en-US/docs/Web/API/Node/nodeType

They use such naming 
`Node.ELEMENT_NODE (1)` 

and ... ``Node.ATTRIBUTE_NODE``

so `element` and `node` can be used together  to describe exactly what type of Node is required.

PR #404, opened 2023-01-07
Add _WD_GetFreePort

Danp2:
### Proposed changes

Add function _WD_GetFreePort

### Checklist

### Types of changes

- [ ] Other (please describe)

No easy means to identify an alternate port number that is available for use

### What is the new behavior?

_WD_GetFreePort can be called to obtain a free port number within a defined range

### System under test

- OS: [e.g. Windows 10]
- OS Arch.: [e.g. X64]
- Browser [e.g. firefox]

### DIFF
--- a/wd_helper.au3
+++ b/wd_helper.au3
@@ -2910,9 +2910,9 @@ EndFunc   ;==>_WD_JsonActionKey
 ; #FUNCTION# ====================================================================================================================
 ; Name ..........: _WD_GetFreePort
 ; Description ...:  Locate and return an available TCP port within a defined range
-; Syntax ........: _WD_GetFreePort($iMinPort,  $iMaxPort)
-; Parameters ....: $iMinPort            - Starting port number
-;                  $iMaxPort            - Ending port number
+; Syntax ........: _WD_GetFreePort([$iMinPort = Default[,  $iMaxPort = Default]])
+; Parameters ....: $iMinPort - [optional] Starting port number. Default is 64000
+;                  $iMaxPort - [optional] Ending port number. Default is $iMinPort or 65000
 ; Return values .: Success - Available TCP port number
 ;                  Failure - 0 and @error set to $_WD_ERROR_NotFound 
 ; Author ........: Danp2
@@ -2922,9 +2922,12 @@ EndFunc   ;==>_WD_JsonActionKey
 ; Link ..........:
 ; Example .......: No
 ; ===============================================================================================================================
-Func _WD_GetFreePort($iMinPort, $iMaxPort)
+Func _WD_GetFreePort($iMinPort = Default, $iMaxPort = Default)
 	Local Const $sFuncName = "_WD_GetFreePort"
 	Local $iResult = 0, $iErr = $_WD_ERROR_NotFound
+
+	If $iMaxPort = Default Then $iMaxPort = ($iMinPort = Default) ? 65000 : $iMinPort
+	If $iMinPort = Default Then $iMinPort = 64000
 	Local $aPorts = __WinAPI_GetTcpTable()
 
 	If Not @error Then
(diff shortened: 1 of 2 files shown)


wd_helper.au3
```diff
@@ -2906,6 +2906,71 @@ Func _WD_JsonActionKey($sType, $sKey, $iSuffix = Default)
...
+	Local $aPorts = __WinAPI_GetTcpTable()
+
+	If Not @error Then
+		For $iPort = $iMinPort To $iMaxPort
+			_ArraySearch($aPorts, $iPort, Default, Default, Default, Default, Default, 3)
+			If @error = 6 Then Return $iPort
+		Next
+	Endif
+
+	Return 0
```
mlipok, line 2935: What about reporting `@error` ?

mlipok:
I also start thinking about:
Should we check if port providen by user is free or used and return specific error instead runing webdriver exe to check what will happen ?

wd_helper.au3
```diff
@@ -2906,6 +2906,71 @@ Func _WD_JsonActionKey($sType, $sKey, $iSuffix = Default)
...
+EndFunc   ;==>_WD_GetFreePort
+
+Func __WinAPI_GetTcpTable()
+    ;funkey 2012.12.14
+	;https://www.autoitscript.com/forum/topic/146671-getextendedtcptable-get-netstat-information/?tab=comments#comment-1038649
```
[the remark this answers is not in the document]
Danp2: It's a tab vs spaces issue. I'll fix it.

wd_helper.au3
Danp2: I'll update to use __WD_Error like other functions.

Danp2:
> Should we check if port providen by user is free or used and return specific error instead runing webdriver exe to check what will happen ?

IDK. I guess that could be a nice option to add. I would think you could do something like this --
```autoit
If Not _WD_GetFreePort(9515) Then
    ; Take action here due to unavailable port
EndIf
```

However, this is problematic because you would likely want to add this check in `_WD_Startup` and `_WD_GetFreePort` doesn't exist in wd_core.au3. I specifically placed it in wd_helper.au3 because it isn't "core" webdriver functionality IMO.

wd_helper.au3
```diff
@@ -2922,9 +2922,12 @@ EndFunc   ;==>_WD_JsonActionKey
 ; Link ..........:
 ; Example .......: No
 ; ===============================================================================================================================
-Func _WD_GetFreePort($iMinPort, $iMaxPort)
+Func _WD_GetFreePort($iMinPort = Default, $iMaxPort = Default)
 	Local Const $sFuncName = "_WD_GetFreePort"
 	Local $iResult = 0, $iErr = $_WD_ERROR_NotFound
+
+	If $iMaxPort = Default Then $iMaxPort = ($iMinPort = Default) ? 65000 : $iMinPort
+	If $iMinPort = Default Then $iMinPort = 64000
```
mlipok, line 2930: I propose to defult values
```autoit
If $iMinPort = Default Then $iMinPort = $_WD_PORT
If $iMaxPort = Default Then $iMaxPort = ($iMaxPort = Default) ? ($iMinPort + 100) : $iMaxPort
```

Danp2: @mlipok Appreciate the feedback. However, I have some concerns regarding your proposed changes --

- This function will be used in other areas (such as BiDi), so it is intended to be generic in nature and not specifically related to the port used for web driver communication

- If $_WD_Port <> 0, then that means the user has already assigned a value using _WD_Option. It makes more sense to me if these were the order of events --
```
1. Obtain a port number using _WD_GetFreePort (optional)
2. Set the port number using _WD_Option
3. Call _WD_Startup
```
- I don't see how your 2nd line modifying $iMaxPort can even function properly. Care to explain?

mlipok: Ok.
Should we also add som usage to _wd_demo.au3 ?

mlipok:
Should we add some example usage to wd_demo.au3 ?

Danp2:
I don't think so. We can't cover every possible option in wd_demo. I think this could be a good FAQ for the wiki.

PR #405, opened 2023-01-27
UserTesting: Add ability to read code from file

Danp2:
### Proposed changes

Describe the big picture of your changes here to communicate to the maintainers why we should accept this pull request.<br>
Please ensure you have read and noticed the checklist below.

### Checklist

Put an `x` in the boxes that apply. If you're unsure about any of them, don't hesitate to ask. We are here to help!<br>
This is simply a reminder of what we are going to look for before merging your code.

- [x] I have read and noticed the [CODE OF CONDUCT](https://github.com/Danp2/au3WebDriver/blob/master/docs/CODE_OF_CONDUCT.md) document
- [x] I have read and noticed the [CONTRIBUTING](https://github.com/Danp2/au3WebDriver/blob/master/docs/CONTRIBUTING.md) document
- [ ] I have added necessary documentation or screenshots (if appropriate)

### Types of changes

Please check `x` the type of change your PR introduces:

- [ ] Bugfix (change which fixes an issue)
- [x] Feature (change which adds functionality)
- [ ] Code style update (formatting, renaming)
- [x] Refactoring (functional, structural)
- [ ] Documentation content changes
- [ ] Other (please describe)

### What is the current behavior?

wd_demo must be exited and restarted to test changes to `UserTesting` code

### What is the new behavior?

`UserTesting` now reads code from a file (usertesting.txt), which allows for code modification / execution without exiting wd_demo.

### Additional context

Support for embedded code still exists and can be enabled by setting `$bStyle` equal to false

### System under test

Please complete the following information.

- OS: [e.g. Windows 10]
- OS Arch.: [e.g. X64]
- Browser [e.g. firefox]
- Browser version [e.g. 96.0.3]

PR #407, opened 2023-01-29
usertesting() modified usertesting.txt concept

mlipok:
### Proposed changes

modified usertesting.txt concept

### Checklist

### Types of changes

- [ ] Other (please describe)

works ok

### What is the new behavior?

easier to say "Just modify " 
```autoit
Func __UserTesting_2()
....
EndFunc   ;==>__UserTesting_2
```
and send it back to us that we will be able to check your case.

### Additional context

none

### System under test

not related

### DIFF
--- a/usertesting.txt
+++ b/usertesting.txt
@@ -1,0 +1,11 @@
+	_WD_Navigate($sSession, 'https://www.google.com')
+	_WD_LoadWait($sSession)
+
+	ConsoleWrite("- Test 1:" & @CRLF)
+	_WD_NewTab($sSession, False, Default, "https://www.chrome.com", "noreferrer")
+	_WD_LoadWait($sSession, 1000)
+	_WD_Attach($sSession, "Chrome")
+	_WD_FindElement($sSession, $_WD_LOCATOR_ByXPath, '//input[@title='Search']')
+
+	ConsoleWrite("- Test 2:" & @CRLF)
+	_WD_WaitElement($sSession, $_WD_LOCATOR_ByCSSSelector, '#fake', 1000, 3000, $_WD_OPTION_NoMatch)
--- a/wd_demo.au3
+++ b/wd_demo.au3
@@ -1083,16 +1083,27 @@ Func DemoStyles()
 EndFunc   ;==>DemoStyles
 
 #Region - UserTesting
-Func UserTesting() ; here you can replace the code to test your stuff before you ask on the forum
-	_WD_Navigate($sSession, 'https://www.google.com')
-	_WD_LoadWait($sSession, 10, Default, Default, $_WD_READYSTATE_Interactive)
+Func UserTesting()
+	Local $bStyle = True ; Change to false to use older style
 
-	ConsoleWrite("- Test 1:" & @CRLF)
-	_WD_FindElement($sSession, $_WD_LOCATOR_ByXPath, '')
+	If $bStyle Then ; New style
+		; Modify the contents of usertesting.txt to change the code being executed.
+		; Changes can be made and executed without restarting this script
+		Local $aCmds = FileReadToArray("usertesting.txt")
 
-	ConsoleWrite("- Test 2:" & @CRLF)
-	_WD_WaitElement($sSession, $_WD_LOCATOR_ByCSSSelector, '#fake', 1000, 3000, $_WD_OPTION_NoMatch)
-;~ 	Exit
+		For $sCmd In $aCmds
+			Execute($sCmd)
+		Next
+	Else ; Old style
+		_WD_Navigate($sSession, 'https://www.google.com')
+		_WD_LoadWait($sSession, 10, Default, Default, $_WD_READYSTATE_Interactive)
+
+		ConsoleWrite("- Test 1:" & @CRLF)
+		_WD_FindElement($sSession, $_WD_LOCATOR_ByXPath, '')
+
+		ConsoleWrite("- Test 2:" & @CRLF)
+		_WD_WaitElement($sSession, $_WD_LOCATOR_ByCSSSelector, '#fake', 1000, 3000, $_WD_OPTION_NoMatch)
+	EndIf	
 EndFunc   ;==>UserTesting
 
 ; if necessary, add any additional function required for testing within this region here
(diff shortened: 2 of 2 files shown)


mlipok:
please change destination repository to:
https://github.com/Danp2/au3WebDriver/tree/demo_usertesting

Danp2:
This forces the Yes/No selection each time you want to run the UserTesting routine, so do think I can agree with this. The point of #405 is to speed up the testing process, and forcing the user to make a selection of each run is counterproductive.

mlipok:
> This forces the Yes/No selection each time you want to run the UserTesting routine, so do think I can agree with this. 

We can change the decision to option (checkbox) in wd_demo.au3 gui.
What you think abou such solution ?

> The point of #405 is to speed up the testing process, and forcing the user to make a selection of each run is counterproductive.

The option on the gui (checkbox) will solve this problem.
We can even give an addtitional line to comment/commentout to make the option default

Danp2:
>  What is your opinion on such change (adding checkbox on GUI)

I'm open to this, but I'm not convinced that a checkbox is the best solution. Have you considered other options? For example, you could use a context menu to switch between the two execution types. This is how I did it in my BiDi testing script --

```autoit
	Local $iShow = GUICtrlCreateButton("Show Maps", 95, 75, 75, 25)
	Local $iMenu2 = GUICtrlCreateContextMenu($iShow)
	Local $iEvents = GUICtrlCreateMenuItem("Events", $iMenu2, Default, 1)
	Local $iResults = GUICtrlCreateMenuItem("Results", $iMenu2, Default, 1)
	GUICtrlSetState($iEvents, $GUI_CHECKED)
```

mlipok:
Will continue in the comming week.

mlipok:
We made it a bit too complicated.
I have a simpler solution in mind.
I will prepare an appropriate change that will show what I mean.

mlipok:
take a look now

mlipok:
@Danp2 hope you keep your eyes on this ?

Danp2:
@mlipok It would be beneficial IMO if there was some distinction between `UserTesting1` and `UserTesting2`. My suggestion would be to keep `UserTesting` and rename the other so that it better describes the situation (ie: `DynamicTesting` or something of that nature).

mlipok:
FuncTesting
FileTesting

mlipok:
Or we should stay with 
UserTesting

and only add
FileTesting

mlipok:
User file sounds great

Danp2:
@mlipok Changes have been committed to #405 to split the "user" functions into two separate functions and rename the file. Please review your other changes here and then decide if this PR needs to still be updated / merged.

mlipok:
> @mlipok Changes have been committed to #405 to split the "user" functions into two separate functions and rename the file. Please review your other changes here and then decide if this PR needs to still be updated / merged.

first comment:
`UserTesting.txt` should be renamed to `UserFile_0.au3`
It must have *.au3 extenstion to still support highlighting when you create, and maintain the file.

Danp2:
> `UserTesting.txt` should be renamed to `UserFile_0.au3` It must have *.au3 extension to still support highlighting when you create, and maintain the file.

Please reread my earlier post as this file has already been renamed.

mlipok:
> Please reread my earlier post as this file has already been renamed.

Ah I see. Thanks.

mlipok:
**comment no.2:**
I propose to rename the file from:  `usertesting.au3` to `FileTesting.au3` to keep consistency with function name

**comment no.2:**
I propose to use `FileOpenDialog` to have possibility to maintain many scenarios at onece by having possiblity to select file to check.
You can say to forum members: `Please provide your own "FileTesting_NickName.au3" to make it easy for us to test your case.`

**comment no.3:**
IMO my version has better error handling

**comment no.4:**
`UserFile()` should not be within `UserTesting region` 

because of this following comment:
```autoit
; if necessary, add any additional function required for testing within this region here
#EndRegion - UserTesting
```
is not valid in relation `UserFile()` I mean `UserFile()`  should not be modified in any case but it can use any special user function added within 
```autoit
#Region - UserTesting
....
#EndRegion - UserTesting
```

I'm thinking about scenario when some advance user want to test something and share the code to community.
But the code requries some additionall functions which can not be placed in `UserFile_NickName_v1.au3`
 or `UserFile_NickName_v2.au3` etc....

in such scenario we should expect to get modified
```autoit
#Region - UserTesting
....
#EndRegion - UserTesting
```
and additional files `UserFile_NickName_v1.au3` , `UserFile_NickName_v2.au3` etc....

Danp2:
>  I propose to rename the file from: usertesting.au3 to FileTesting.au3 to keep consistency with function name

The function name is `UserFile`, so your propose doesn't add consistency.

>  IMO my version has better error handling

No disagreement there.

>  UserFile() should not be within UserTesting region

Maybe. My view was that this region would encompass all "user testing" functions instead of only the `userTesting` function. Honestly, I don't have strong feeling either way about this, especially when you consider that wd_demo wasn't originally intended to be used in this manner.

mlipok:
> > IMO my version has better error handling
> 
> No disagreement there.

Do not know what you mean in your statement here.

Danp2:
@mlipok It essentially means "I agree". 😄

mlipok:
> I don't have strong feeling either way about this, especially when you consider that wd_demo wasn't originally intended to be used in this manner.

Yes. but it becames also a helper tool to test user scripts, and provide support to user cases.

As I understand it, you also have this belief since you yourself proposed the possibility of loading the user's file, somewhat on the runtime.

mlipok:
> the possibility of loading the user's file, somewhat on the runtime

btw. awesom concept which I also plan to provide in my own testing modes in all of my programs/scirpts, to have the possiblity to test code on the fly at runtime.

Danp2:
>  I propose to rename the file from: usertesting.au3 to UserFile.au3 to keep consistency with function name

Yes, that would add consistency to the naming. However, I'm not convinced that it improves usability / understanding for the end user. I suggest leaving the name as-is, and we should add a header to the file explaining what it is, how to use it, which function in wd_demo it is associated with, etc.

mlipok:
> . I suggest leaving the name as-is, and we should add a header to the file explaining what it is, how to use it, which function in wd_demo it is associated with, etc.

If you agree.. merge my PR to yours and make further needed changes.
Then I finally review your PR and job will be done.

Issue #403, opened 2022-12-16
_WD_ElementActionEx fails if element isn't visible in viewport


Danp2:
### Describe the bug

_WD_ElementActionEx doesn't scroll the target element into view, which can lead to failure of some webdriver actions.

### How to reproduce

Run the following in wd_demo's UserTesting function --

```autoit
Func UserTesting() ; here you can replace the code to test your stuff before you ask on the forum
	Local $sURL = "https://about.google/"
	Local $sXpath = "(//a[normalize-space()='View all our commitments'])[1]"

	_WD_Navigate ($sSession, $sURL)
	_WD_LoadWait ($sSession)

	$sElement = _WD_FindElement($sSession, $_WD_LOCATOR_ByXPath, $sXpath)
	_WD_ElementActionEx($sSession, $sElement, "hover")
EndFunc   ;==>UserTesting
```

Here's the corresponding logs --

```
    __WD_Post: URL=HTTP://127.0.0.1:4444/session/1be1e01e-0774-40c8-b116-bb7e3c686e3d/actions; Data={"actions":[{"id":"hover","type":"pointer","parameters":{"pointerType":"mouse"},"actions":[{"type":"pointerMove","duration":100,"x":0,"y":0,"origin":{"ELEMENT":"dd776eaa-a915-4332-b3ca-532985074fe3","element-6066-11e4-a52e-4f735466cecf":"dd776eaa-a915-4332-b3ca-532985074fe3"}}]}]}
[...]

PR #421, opened 2023-02-22
_WD_ElementActionEx - inner comments suplemented

mlipok:
### Proposed changes

Sleep() should not be used directly to have the $_WD_ERROR_UserAbort feature available

### Checklist

### Types of changes

- [ ] Other (please describe)

after scrolling script is stopped by Sleep() function whith out possibility to interact with.

### What is the new behavior?
__WD_Sleep() gives more interaction to the script

### Additional context
I would like to do this more flexible the script should check if it is visible or not and then scroll or not, after which should wait to be visible, and not only simply Sleep()

### System under test
not related

### DIFF
--- a/wd_helper.au3
+++ b/wd_helper.au3
@@ -2595,19 +2595,23 @@ Func _WD_ElementActionEx($sSession, $sElement, $sCommand, $iXOffset = Default, $
 
 	If $bScrollView Then
 		_WD_ExecuteScript($sSession, "arguments[0].scrollIntoView(false);", __WD_JsonElement($sElement))
-		Sleep(500)
+		__WD_Sleep(500)
 	EndIf
 
-	Switch $iActionType
-		Case 1
-			$sAction = StringFormat($sActionTemplate, $sPreAction, $iXOffset, $iYOffset, $sElement, $sElement, $sPostHoverAction, $sPostAction)
-			$sResult = _WD_Action($sSession, 'actions', $sAction)
-			$iErr = @error
+	If @error Then
+		$iErr = @error
+	Else
+		Switch $iActionType
+			Case 1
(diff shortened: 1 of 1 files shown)


Danp2:
> Sleep() should not be used directly to have the $_WD_ERROR_UserAbort feature available

Generally, I agree with this statement. However, I'm not sure I agree in this instance. If you recall, the code in question came out of the following discussion --

https://github.com/Danp2/au3WebDriver/issues/403#issuecomment-1354970960

The Sleep statement is needed to allow time for the element to scroll into view. I don't see the harm in using Sleep() in this situation because --

- it's only for half a second
- there isn't a loop involved where the sleep would occur multiple times

Danp2:
> I would like to do this more flexible the script should check if it is visible or not and then scroll or not, after which should wait to be visible, and not only simply Sleep()

Yes, this could be done. However, there are several stumbling blocks --

- It would increase complexity
- It would increase execution time

mlipok:
So I will propose a separate function to check and force the visibility of an element on the view through scroll.

mlipok:
> The Sleep statement is needed to allow time for the element to scroll into view. I don't see the harm in using Sleep() in this situation because --
> 
> * it's only for half a second
> * there isn't a loop involved where the sleep would occur multiple times

Ok. I accept your point of view. so I'll just modify this part a bit by adding a comment to the code so that I don't get bogged down in the future.

What you think about ?

Danp2:
> So I will propose a separate function to check and force the visibility of an element on the view through scroll.

Ok

> I'll just modify this part a bit by adding a comment to the code so that I don't get bogged down in the future.

That's fine.

Danp2 (approved):

PR #431, opened 2023-03-02
Add _WD_WaitScript function

ye7iaserag:
### Proposed changes

Add a new function that takes JavaScript code and waits until the code returns true

### Checklist

### Types of changes

- [ ] Other (please describe)

New function

### What is the new behavior?

New function

### System under test

- OS: [e.g. Windows 10]
- OS Arch.: [e.g. X64]
- Browser [e.g. chrome]
- Browser version [e.g. 110.0]

### DIFF
--- a/wd_helper.au3
+++ b/wd_helper.au3
@@ -457,8 +457,8 @@ EndFunc   ;==>_WD_WaitElement
 ;                  $sJavaScript - JavaScript to run
 ;                  $iDelay    - [optional] Milliseconds to wait before initially checking status
 ;                  $iTimeout  - [optional] Period of time (in milliseconds) to wait before exiting function
-; Return values .: Success - True.
-;                  Failure - "" (empty string) and sets @error to one of the following values:
+; Return values .: Success - True
+;                  Failure - False and sets @error to one of the following values:
 ;                  - $_WD_ERROR_InvalidArgue
 ;                  - $_WD_ERROR_RetValue
 ;                  - $_WD_ERROR_Exception
@@ -489,7 +489,7 @@ Func _WD_WaitScript($sSession, $sJavaScript, $iDelay = Default, $iTimeout = Defa
 		; prevent multiple errors https://github.com/Danp2/au3WebDriver/pull/290#issuecomment-1100707095
 		Local $_WD_DEBUG_Saved = $_WD_DEBUG ; save current DEBUG level
 
-		; Prevent logging from _WD_FindElement if not in Full debug mode
+		; Prevent logging from _WD_ExecuteScript if not in Full debug mode
 		If $_WD_DEBUG <> $_WD_DEBUG_Full Then $_WD_DEBUG = $_WD_DEBUG_None
 
 		Local $hWaitTimer = TimerInit()
--- a/wd_helper.au3
+++ b/wd_helper.au3
@@ -450,6 +450,73 @@ Func _WD_WaitElement($sSession, $sStrategy, $sSelector, $iDelay = Default, $iTim
 EndFunc   ;==>_WD_WaitElement
 
 ; #FUNCTION# ====================================================================================================================
+; Name ..........: _WD_WaitScriptTrue
+; Description ...: Wait for a JavaScript snippet to return true.
+; Syntax ........: _WD_WaitScriptTrue($sSession, $sJavaScript[, $iDelay = Default[, $iTimeout = Default[, $iOptions = Default]]])
+; Parameters ....: $sSession  - Session ID from _WD_CreateSession
+;                  $sJavaScript - JavaScript to run
+;                  $iDelay    - [optional] Milliseconds to wait before initially checking status
+;                  $iTimeout  - [optional] Period of time (in milliseconds) to wait before exiting function
+; Return values .: Success - True.
+;                  Failure - "" (empty string) and sets @error to one of the following values:
+;                  - $_WD_ERROR_Exception
+;                  - $_WD_ERROR_Timeout
+;                  - $_WD_ERROR_UserAbort
+; Author ........: yehiaserag
+; Modified ......:
+; Remarks .......:
+; Related .......: _WD_ExecuteScript
+; Link ..........:
+; Example .......: No
+; ===============================================================================================================================
+Func _WD_WaitScriptTrue($sSession, $sJavaScript, $iDelay = Default, $iTimeout = Default)
+	Local Const $sFuncName = "_WD_WaitCallbackTrue"
+	Local Const $sParameters = 'Parameters:   JavaScript=' & $sJavaScript & '   Delay=' & $iDelay & '   Timeout=' & $iTimeout
+	Local $iErr
+	$_WD_HTTPRESULT = 0
+	$_WD_HTTPRESPONSE = ''
+
+	If $iDelay = Default Then $iDelay = 0
+	If $iTimeout = Default Then $iTimeout = $_WD_DefaultTimeout
+
+	__WD_Sleep($iDelay)
+	$iErr = @error
+
+	; prevent multiple errors https://github.com/Danp2/au3WebDriver/pull/290#issuecomment-1100707095
+	Local $_WD_DEBUG_Saved = $_WD_DEBUG ; save current DEBUG level
+
+	; Prevent logging from _WD_FindElement if not in Full debug mode
+	If $_WD_DEBUG <> $_WD_DEBUG_Full Then $_WD_DEBUG = $_WD_DEBUG_None
+
+	Local $hWaitTimer = TimerInit()
+	While 1
+		If $iErr Then ExitLoop
+
+		$sValue = _WD_ExecuteScript($sSession, $sJavaScript, Default, Default, $_WD_JSON_Value)
+		$iErr = @error
+
+		If $iErr <> $_WD_ERROR_Success Then
+			; Exit loop if unexpected error occurs
+			ExitLoop
+		ElseIf $sValue = False Then
+			If (TimerDiff($hWaitTimer) > $iTimeout) Then
+				$iErr = $_WD_ERROR_Timeout
+				ExitLoop
+			EndIf
+
+			__WD_Sleep(10)
+			$iErr = @error
+		Else
+			$iErr = $_WD_ERROR_Success
+			ExitLoop
+		EndIf
+	WEnd
+	$_WD_DEBUG = $_WD_DEBUG_Saved ; restore DEBUG level
+
+	Return SetError(__WD_Error($sFuncName, $iErr, $sParameters), 0, $sValue)
+EndFunc   ;==>_WD_WaitScriptTrue
+
+; #FUNCTION# ====================================================================================================================
 ; Name ..........: _WD_DebugSwitch
 ; Description ...: Switch to new debug level or switch back to saved debug level
 ; Syntax ........: _WD_DebugSwitch([$vMode = Default])
--- a/wd_helper.au3
+++ b/wd_helper.au3
@@ -466,7 +466,7 @@ EndFunc   ;==>_WD_WaitElement
 ;                  - $_WD_ERROR_UserAbort
 ; Author ........: yehiaserag
 ; Modified ......:
-; Remarks .......:
+; Remarks .......: The Javascript needs to return either True or False.
 ; Related .......: _WD_ExecuteScript
 ; Link ..........:
 ; Example .......: No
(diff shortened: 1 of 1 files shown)


Danp2:
@ye7iaserag Thanks for your submission. A few questions for you --

- Can you provide a short code snippet (ie: something that can be used in wd_demo's UserTesting function) that demonstrates the proper usage of this function?
- Should the function perform any validation of the provided Javascript? For example, should an error be returned if the JS doesn't begin with a "return"?

Danp2:
Another thought would be to modify your submission as follows --

- Rename the function to `_WD_ExecuteScriptEx`, since it is a wrapper for _WD_ExecuteScript
- Add an optional parameter `$vWaitReturn`, which should default to Default
- When `$vWaitReturn` isn't equal to Default, perform the comparison check on the scripts return value against `$vWaitReturn`, which allows for more comparisons than just `true`.

IMO, these changes would make your function more dynamic / useful to the user.

ye7iaserag:
- Regarding the code snippet no worries, I'm on it.
- Validation: You are right maybe a simple `StringRight` call can be used here.
- Regarding function name, I tried to be as close to the other libs like `puppeteer` that uses `waitForFunction` for such case, wile I do not think `_WD_ExecuteScriptEx` would reflect what the function is actuallu doing, it's close in functionaly to `_WD_WaitElement`, so I followed the same convention.
- I don't think `$vWaitReturn` would extend existing functionality, since you can write any condition/expression you want in the JS script, any thing that you would add in `$vWaitReturn` can be added directly in script.

Please let me know what you think of the above points.

Danp2:
Hi @ye7iaserag,

The current implementation only works with a True/False result from the Javascript, and "success" is only achieved when True is returned. I was suggesting that the function could be more powerful if it allowed other return values to indicate success. For example --

- Loop until the result is False instead of True
- Loop until result is a specific numeric value
- Etc

Let me know if this helps explain why I suggested adding `$vWaitReturn`. 

Regards, Dan

ye7iaserag:
But you can manipulate that in the JS expression, still don't get your point

an example of usage is:
```
_WD_WaitScriptTrue($sSession, 'return window.getComputedStyle(document.querySelector("div#div1")).getPropertyValue("opacity") === "0" || document.querySelector("div#div2").innerText === "Completed!"')
```

and in the same way you can achieve any required result by manipulating the expression 
an example would be changing the above code to be:
```
```
_WD_WaitScriptTrue($sSession, 'return window.getComputedStyle(document.querySelector("div#div1")).getPropertyValue("opacity") === "0.5"')
```
Let me know if we are still not on the same page

**Comment by Danp2** [comment]
OIC. That makes sense now. Thanks for clarifying.

**Comment by ye7iaserag** [comment]
I agree on both points

**Comment by ye7iaserag** [comment]
I think I have to remove the check for the `return` string on the left of script.
One use cases that I had going was something like
```
let name = document.querySelector(".success");if(!!name){return name.innerText.includes("' & $usernameValue & '")}else{return false;}
```
Now this fails

**Comment by Danp2** [comment]
@ye7iaserag It's fine if you want to remove that check.

File: wd_helper.au3 [hunk 1]
```diff
@@ -449,6 +449,78 @@ Func _WD_WaitElement($sSession, $sStrategy, $sSelector, $iDelay = Default, $iTim
...
+; Remarks .......:
+; Related .......: _WD_ExecuteScript
+; Link ..........:
+; Example .......: No
+; ===============================================================================================================================
+Func _WD_WaitScript($sSession, $sJavaScript, $iDelay = Default, $iTimeout = Default)
+	Local Const $sFuncName = "_WD_WaitScript"
+	Local Const $sParameters = 'Parameters:   JavaScript=' & $sJavaScript & '   Delay=' & $iDelay & '   Timeout=' & $iTimeout
+	Local $iErr
+	Local $sValue = False
```
Danp2, line 478: This should really be `$bValue` because it will hold a boolean value.

wd_helper.au3
```diff
@@ -449,6 +449,78 @@ Func _WD_WaitElement($sSession, $sStrategy, $sSelector, $iDelay = Default, $iTim
...
+; Return values .: Success - True.
+;                  Failure - "" (empty string) and sets @error to one of the following values:
+;                  - $_WD_ERROR_InvalidArgue
+;                  - $_WD_ERROR_RetValue
+;                  - $_WD_ERROR_Exception
+;                  - $_WD_ERROR_Timeout
+;                  - $_WD_ERROR_UserAbort
+; Author ........: yehiaserag
+; Modified ......:
+; Remarks .......:
```
Danp2, line 469: Please add a remark here to indicate that the Javascript needs to return either True or False.

Danp2 (changes requested):

wd_helper.au3
```diff
@@ -475,7 +475,7 @@ Func _WD_WaitScript($sSession, $sJavaScript, $iDelay = Default, $iTimeout = Defa
 	Local Const $sFuncName = "_WD_WaitScript"
 	Local Const $sParameters = 'Parameters:   JavaScript=' & $sJavaScript & '   Delay=' & $iDelay & '   Timeout=' & $iTimeout
 	Local $iErr
-	Local $sValue = False
+	Local $bValue = False
```
mlipok, line 478: it is too common name for variable especially in the case when you are in fact not trying to get value but to test

please rename it to `$bResult` or `$bTest`

wd_helper.au3
```diff
@@ -449,6 +449,78 @@ Func _WD_WaitElement($sSession, $sStrategy, $sSelector, $iDelay = Default, $iTim
...
+; #FUNCTION# ====================================================================================================================
+; Name ..........: _WD_WaitScript
+; Description ...: Wait for a JavaScript snippet to return true.
+; Syntax ........: _WD_WaitScript($sSession, $sJavaScript[, $iDelay = Default[, $iTimeout = Default[, $iOptions = Default]]])
+; Parameters ....: $sSession  - Session ID from _WD_CreateSession
+;                  $sJavaScript - JavaScript to run
+;                  $iDelay    - [optional] Milliseconds to wait before initially checking status
+;                  $iTimeout  - [optional] Period of time (in milliseconds) to wait before exiting function
+; Return values .: Success - True.
+;                  Failure - "" (empty string) and sets @error to one of the following values:
```
mlipok, line 461: not empty string
```autoit
Failure - False and sets @error to one of the following values:
```

wd_helper.au3
```diff
@@ -449,6 +449,78 @@ Func _WD_WaitElement($sSession, $sStrategy, $sSelector, $iDelay = Default, $iTim
...
+	If StringLeft($sJavaScript, 6) <> "return" Then
+		$iErr = $_WD_ERROR_InvalidArgue
+	Else
+		__WD_Sleep($iDelay)
+		$iErr = @error
+
+		; prevent multiple errors https://github.com/Danp2/au3WebDriver/pull/290#issuecomment-1100707095
+		Local $_WD_DEBUG_Saved = $_WD_DEBUG ; save current DEBUG level
+
+		; Prevent logging from _WD_FindElement if not in Full debug mode
```
mlipok, line 492: `; Prevent logging from _WD_ExecuteScript if not in Full debug mode`

wd_helper.au3
```diff
@@ -449,6 +449,78 @@ Func _WD_WaitElement($sSession, $sStrategy, $sSelector, $iDelay = Default, $iTim
...
+				; Exit loop if unexpected error occurs
+				ExitLoop
+			ElseIf $bValue = False Then
+				If (TimerDiff($hWaitTimer) > $iTimeout) Then
+					$iErr = $_WD_ERROR_Timeout
+					ExitLoop
+				EndIf
+
+				__WD_Sleep(10)
+				$iErr = @error
```
mlipok, line 512: if err then exitloop ?

wd_helper.au3
Danp2: @mlipok I think you are being too picky here. The variable is declared locally, so I don't see how giving it a common name could cause any issues. Technically, `true` and `false` *are* values. 😉 

Please elaborate if you disagree.

wd_helper.au3
Danp2: Good catch

wd_helper.au3
Danp2: Shouldn't be needed due to the check at the beginning of the loop.

mlipok: IMO in such case `Abort` will not be possible

wd_helper.au3
```diff
@@ -449,6 +449,78 @@ Func _WD_WaitElement($sSession, $sStrategy, $sSelector, $iDelay = Default, $iTim
...
+Func _WD_WaitScript($sSession, $sJavaScript, $iDelay = Default, $iTimeout = Default)
+	Local Const $sFuncName = "_WD_WaitScript"
+	Local Const $sParameters = 'Parameters:   JavaScript=' & $sJavaScript & '   Delay=' & $iDelay & '   Timeout=' & $iTimeout
+	Local $iErr
+	Local $bValue = False
+
+	If $iDelay = Default Then $iDelay = 0
+	If $iTimeout = Default Then $iTimeout = $_WD_DefaultTimeout
+
+	If StringLeft($sJavaScript, 6) <> "return" Then
```
Danp2, line 483: This still needs to be modified or removed.

wd_helper.au3
```diff
@@ -449,6 +449,78 @@ Func _WD_WaitElement($sSession, $sStrategy, $sSelector, $iDelay = Default, $iTim
...
+		Local $_WD_DEBUG_Saved = $_WD_DEBUG ; save current DEBUG level
+
+		; Prevent logging from _WD_ExecuteScript if not in Full debug mode
+		If $_WD_DEBUG <> $_WD_DEBUG_Full Then $_WD_DEBUG = $_WD_DEBUG_None
+
+		Local $hWaitTimer = TimerInit()
+		While 1
+			If $iErr Then ExitLoop
+
+			$bValue = _WD_ExecuteScript($sSession, 'return !!((function(){' & $sJavaScript & '})())', Default, Default, $_WD_JSON_Value)
```
Danp2, line 499: @ye7iaserag Can you explain what your recent change to this line does?

Danp2 (changes requested):

wd_helper.au3
ye7iaserag: You can write any expression in a way that starts with return, the code will just be more complex 
This is why I left it

wd_helper.au3
ye7iaserag: It wraps the code in a IIFE that checks for a truthy value.
In simple terms this creates an immediately invoked fucking that forces the user code to return a boolean even if the code does not return anything

wd_helper.au3
Danp2: Would it be beneficial to switch to StringInStr() instead of StringLeft()?

wd_helper.au3
```diff
@@ -449,6 +449,78 @@ Func _WD_WaitElement($sSession, $sStrategy, $sSelector, $iDelay = Default, $iTim
...
+
+		; prevent multiple errors https://github.com/Danp2/au3WebDriver/pull/290#issuecomment-1100707095
+		Local $_WD_DEBUG_Saved = $_WD_DEBUG ; save current DEBUG level
+
+		; Prevent logging from _WD_ExecuteScript if not in Full debug mode
+		If $_WD_DEBUG <> $_WD_DEBUG_Full Then $_WD_DEBUG = $_WD_DEBUG_None
+
+		Local $hWaitTimer = TimerInit()
+		While 1
+			If $iErr Then ExitLoop
```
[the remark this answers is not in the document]
Danp2: The check is at the beginning so that it will also work for the __WD_Sleep that occurs prior to the start of the loop. 

Seems like we've discussed this previously for one of the other similar functions. 🤔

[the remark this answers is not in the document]
mlipok: Yes I know. I said already that I see that.
I'm not refering to `__WD_Sleep` here.
I'm wondering about refactoring this:
```autoit
While 1
	If $iErr Then ExitLoop
	.....
WEnd
```
to this form
```autoit
Do
	.....
Until $iErr
```

it looks more obvius for me to check `$iErr` on the end not at the begining.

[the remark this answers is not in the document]
mlipok: about the fact that `I'm nitpicky sometimes.`
?
Yes I know.
If so then you can consider this part as resolved.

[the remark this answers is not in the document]
mlipok: > If the loop is restructured to a Do...While, then you would need to also add code so that the loop is never entered if a user abort occurred.

I plan to propose some tweaks (PR) for this.
ASAP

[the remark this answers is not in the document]
Danp2: Why? ☹️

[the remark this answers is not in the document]
mlipok: After rethinking we should leave it here as it is because this following error checking
https://github.com/ye7iaserag/au3WebDriver/blob/master/wd_helper.au3#L496-L497

should be at start to catchup this error:
https://github.com/ye7iaserag/au3WebDriver/blob/master/wd_helper.au3#L486-L487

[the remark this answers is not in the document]
Danp2: Glad we are finally in agreement on this. 😄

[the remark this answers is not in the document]
mlipok: this part should be considered as resolved.

wd_helper.au3
ye7iaserag: You still can't guarantee that all code paths will return in that way...

Danp2 (approved):

PR #433, opened 2023-03-06
_WD_LocateElement - new function

mlipok:
### Proposed changes

New function to lacate element In the interior of all frames

### Checklist

- [x] I have read and noticed the [CODE OF CONDUCT](https://github.com/Danp2/au3WebDriver/blob/master/docs/CODE_OF_CONDUCT.md) document
- [x] I have read and noticed the [CONTRIBUTING](https://github.com/Danp2/au3WebDriver/blob/master/docs/CONTRIBUTING.md) document
- [ ] I have added necessary documentation or screenshots (if appropriate)

### Types of changes

Please check `x` the type of change your PR introduces:

- [ ] Bugfix (change which fixes an issue)
- [x] Feature (change which adds functionality)
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (functional, structural)
- [ ] Documentation content changes
- [ ] Other (please describe)

### What is the current behavior?
Locating the element can be difficult, especially if the page contains many frames, especially if the document structure is nested.

### What is the new behavior?
Easy to locate desired frame that contains matched element.

### Additional context
none

### System under test
not related

Issue #397, opened 2022-08-31
_WD_FrameList - not always takes all frames


mlipok:
### Describe the bug
`_WD_FrameList()` gets list of frames which is not always complete.
It is due to the fact that Not all frames document are load completly.

### How to reproduce
use this following code.
```autoit
Func UserTesting() ; here you can replace the code to test your stuff before you ask on the forum
	_WD_Navigate($sSession, 'https://www.w3schools.com/tags/tryit.asp?filename=tryhtml_iframe')
	_WD_LoadWait($sSession)

	; Example 1 - from 'https://www.w3schools.com' get frame list as string
	Local $sResult = _WD_FrameList($sSession, False)
	#forceref $sResult
	ConsoleWrite("! ---> @error=" & @error & "  @extended=" & @extended & " : Example 1" & @CRLF)
	ConsoleWrite($sResult & @CRLF)

	; Example 2 - from 'https://www.w3schools.com' get frame list as array
	Local $aFrameList = _WD_FrameList($sSession, True)
	ConsoleWrite("! ---> @error=" & @error & "  @extended=" & @extended & " : Example 2" & @CRLF)
	_ArrayDisplay($aFrameList, 'Example 2 - w3schools.com - get frame list as array', 0, 0, Default, 'Absolute Identifiers > _WD_FrameEnter|Relative Identifiers > _WD_FrameEnter|IFRAME attributes|URL|Body ElementID|IsHidden')

	; Example 2 - from 'https://www.w3schools.com' get frame list as array
	$aFrameList = _WD_FrameList($sSession, True)
	ConsoleWrite("! ---> @error=" & @error & "  @extended=" & @extended & " : Example 2 - again" & @CRLF)
	_ArrayDisplay($aFrameList, 'Example 2 - w3schools.com - get frame list as array - again', 0, 0, Default, 'Absolute Identifiers > _WD_FrameEnter|Relative Identifiers > _WD_FrameEnter|IFRAME attributes|URL|Body ElementID|IsHidden')
EndFunc
```
When first `_ArrayDisplay` is displayed wait at least 1 minute before you close the displayed array.
Compare second `_ArrayDisplay` result with the frst result.
Also check `window.top.document.querySelectorAll('iframe')` in browser console.

### Expected behavior

Always full list of frames.

### Screenshots

![image](https://user-images.githubusercontent.com/11089482/187792336-b8bdd789-8e2b-4e4b-ab7b-154607e833ad.png)

### Additional context

none

[...]

PR #434, opened 2023-03-07
_WD_FrameList - <frame..> support + refactoring

mlipok:
### Proposed changes

Refactoring for better getting more data

### Checklist

### Types of changes

- [ ] Other (please describe)

getting the list of frames can stop on specific frames and do not process rest of them

### What is the new behavior?
continue processing frames even if one of them fires any kind of issue.

### Additional context
https://github.com/Danp2/au3WebDriver/issues/397#issuecomment-1451525079

https://github.com/Danp2/au3WebDriver/pull/433#discussion_r1127519862

### System under test
not related

tested with:
```autoit
Func UserTesting() ; here you can replace the code to test your stuff before you ask on the forum
	
		; go to website
	_WD_Navigate($sSession, 'https://www.tutorialspoint.com/html/html_frames.htm#')
	_WD_LoadWait($sSession)

	; check if document context location is Top Window
	ConsoleWrite("> " & @ScriptLineNumber & " IsWindowTop = " & _WD_IsWindowTop($sSession) & @CRLF)

	MsgBox($MB_TOPMOST, "", 'Before checking location of multiple elements on multiple frames' & @CRLF & 'Try the same example with and without waiting about 30 seconds in order to see that many frames should be fully loaded, and to check the differences')

	$aFrameList = _WD_FrameList($sSession, True)
	ConsoleWrite("! ---> @error=" & @error & "  @extended=" & @extended & " : Example : Testing element location in frame set" & @CRLF)
	_ArrayDisplay($aFrameList, 'www.tutorialspoint.com - get frame list as array', 0, 0, Default, $sArrayHeader)
EndFunc
```


wd_helper.au3
```diff
@@ -77,13 +77,14 @@ Global Enum _
...
 
-Global Enum _
+Global Enum _ ; _WD_FrameList() , _WD_LocateElement()
 		$_WD_FRAMELIST_Absolute = 0, _
 		$_WD_FRAMELIST_Relative = 1, _
 		$_WD_FRAMELIST_Attributes = 2, _
 		$_WD_FRAMELIST_URL = 3, _
 		$_WD_FRAMELIST_BodyID = 4, _
 		$_WD_FRAMELIST_FrameVisibility = 5, _
+		$_WD_FRAMELIST_MatchedElements = 6, _ ; array of matched element from _WD_LocateElement()
```
mlipok, line 87: i had it already added for `_WD_LocateElement()`
Please keep it here.

wd_helper.au3
```diff
@@ -875,49 +874,60 @@ Func __WD_FrameList_Internal($sSession, $sLevel, $sFrameAttributes, $bIsHidden,
...
 				If @error Then
 					$sMessage = 'Error occurred on "' & $sLevel & '" level when checking URL'
-				Else
-					$vResult = $sLevel & '|' & $sLevel & '|' & $sFrameAttributes & '|' & $s_URL & '|' & $sCurrentBody_ElementID & '|' & $bIsHidden & @CRLF
 				EndIf
 			EndIf
 		EndIf
 	EndIf
+	$vResult = $sLevel & '|' & $sLevel & '|' & $sFrameAttributes & '|' & $s_URL & '|' & $sCurrentBody_ElementID & '|' & $bIsHidden & '|' & @CRLF
+	ConsoleWrite("> " & $vResult & @CRLF) ; for testing only should be removed before merge.
```
mlipok, line 882: all lines with `; for testing only should be removed before merge.` comment should be removed before merge.

mlipok: Waiting for yor review...

Danp2: Need to update `$sArrayHeader` in wd_demo

Danp2: Also seeing errors generated from line 893 that look like this --

```
    __WD_Post: URL=HTTP://127.0.0.1:4444/session/7ec4e254-b880-4e63-a506-c3996c51ccd4/execute/sync; Data={"script":"return document.querySelectorAll('iframe')[0].outerHTML;", "args":[]}
    __WD_Post ==> Javascript Exception [23] : HTTP status = 500 ResponseText={"value":{"error":"javascript error","message":"TypeError: document.querySelectorAll(...)[0] is undefined","stacktrace":"@https://www.tutorialspoint.com/html/src/vertical_distribution_of_frames.htm:2:23\n@https://www.tutorialspoint.com/html/src/vertical_distribution_of_frames.htm:3:8\n"}}
    _WD_ExecuteScript ==> Javascript Exception [23] : Error occurred when trying to ExecuteScript
    _WD_DebugSwitch:  /  stack size: 1
```

Danp2: `$sMessage` contents gets overwritten by subsequent passes through For loop.

mlipok: > Need to update `$sArrayHeader` in wd_demo

Have it done already in my _WD_Locate testing script so I just add this header here ASAP .

Danp2: > > Need to update `$sArrayHeader` in wd_demo
> 
> Have it done already in my _WD_Locate testing script so I just add this header here ASAP .

It still needs to be updated in wd_demo, correct?

mlipok: Yes it was waiting to be updated.
And ... no as I just update this here: [$sArrayHeader suplemented](https://github.com/Danp2/au3WebDriver/pull/434/commits/b7e6b09efcecf03bdb6fc881cbad57e01f9628c0)

mlipok: > `$sMessage` contents gets overwritten by subsequent passes through For loop.

You mean in all the places where `ContinueLoop` is used just after `$sMessage = ` is set ?

mlipok: I will check what happen when I put message to console like this:
```autoit
$sMessage = 'Error occurred on "' & $sLevel & '" level when trying to check attributes child frames #' & $iFrame
__WD_ConsoleWrite($sFuncName & ": " & $sMessage, $_WD_DEBUG_Info)
ContinueLoop
```

Danp2: > > `$sMessage` contents gets overwritten by subsequent passes through For loop.
> 
> You mean in all the places where `ContinueLoop` is used just after `$sMessage = ` is set ?

Yes

mlipok: Please take a look on this: [$sMessage > __WD_ConsoleWrite](https://github.com/Danp2/au3WebDriver/pull/434/commits/70d168bcc5ef95b45262c2db198a14075bece9c8)

mlipok: > TypeError: document.querySelectorAll(...)[0] is undefined

This is a situation when frames disapers. For example, an ad on a page served and closed.

mlipok: > > TypeError: document.querySelectorAll(...)[0] is undefined
> 
> This is a situation when frames disapers. For example, an ad on a page served and closed. We should treat it as standard behavior.

For this reason I placed an msgbox 
```autoit
MsgBox($MB_TOPMOST, "", 'Before checking location of multiple elements on multiple frames' & @CRLF & 'Try the same example with and without waiting about 30 seconds in order to see that many frames should be fully loaded, and to check the differences')
```

to give user time manually check and be sure that webpage is completly loaded.

mlipok: have some findigs.

```
__WD_Post: URL=HTTP://127.0.0.1:4444/session/e4b327f4-3c2a-4292-affa-d93f56bac169/execute/sync; Data={"script":"return document.querySelectorAll('iframe')[0].outerHTML;", "args":[]}
__WD_Post ==> Javascript Exception [23] : HTTP status = 500 ResponseText={"value":{"error":"javascript error","message":"TypeError: document.querySelectorAll(...)[0] is undefined","stacktrace":"@https://www.tutorialspoint.com/html/src/creating_frames.htm:2:23\n@https://www.tutorialspoint.com/html/src/creating_frames.htm:3:8\n"}}
_WD_ExecuteScript ==> Javascript Exception [23] : Error occurred when trying to ExecuteScript
- Error occurred on "null/0" level when trying to check attributes child frames #0
```

I add some code to get HTML source when this errors occurs.
And I get:
```html
<html><head>
      <title>HTML Frames</title>
   </head>
	
   <frameset rows="10%,80%,10%">
      <frame name="top" src="/html/top_frame.htm">
      <frame name="main" src="/html/main_frame.htm">
      <frame name="bottom" src="/html/bottom_frame.htm">
   
      <noframes>
         <body>
            Your browser does not support frames.
         </body>
      </noframes>
      
   </frameset>
   
</html>
```

so the problem is that this is `frameset` and we will not be able to get:
`return document.querySelectorAll('iframe')[0].outerHTML;`
as this is not `iframe`

so this is bigger problem...

mlipok: I was trying to use:
`return window.frames[" & $iFrame & "].frameElement.outerHTML;`

but have other things to check.

I wonder if `frameset` and `iframe` can be combined in singe `window.document`

mlipok: I was reading here:
https://stackoverflow.com/a/67765623/5314940
is that mean that `frameset` and `iframe` can not be combined together in single document ?

mlipok: Please take a look here: [<frame..> support](https://github.com/Danp2/au3WebDriver/pull/434/commits/a2d9858e2afd09c0cb16a271d174b2f6d4005687)

Danp2: Untested, but I would think that you could do something like this to eliminate multiple calls to _WD_ExecuteScript --

```autoit
$JS = "let nodes = document.querySelectorAll('iframe'); if (nodes) { return nodes[" & $iFrame & "].outerHTML; } else { return window.frames[" & $iFrame & "].frameElement.outerHTML; }"
$sFrameAttributes = _WD_ExecuteScript($sSession, $JS, Default, Default, $_WD_JSON_Value)
```

mlipok: Thanks. 
Will xheck ASAP

mlipok: Works good.
tested with:
```js
function GetFrame(iFrame) {
  let nodes = document.querySelectorAll('iframe');
  if (nodes.length) { return nodes[iFrame].outerHTML; } 
  else { return window.frames[iFrame].frameElement.outerHTML; }
}

GetFrame(0);
```
Please notice `nodes.length` change.

mlipok: Please take a look on:
[$sJavaScript_OuterHTML](https://github.com/Danp2/au3WebDriver/pull/434/commits/24914dc2bc915e5300d698fc6434975483d7482e)

and also on:
[$sArrayHeader + FRAME](https://github.com/Danp2/au3WebDriver/pull/434/commits/57eba9b91a8813901190ba3f7f1bbf0b5ee864f7)

Danp2: Regarding your recent changes --

- Your usage of `StringFormat` in combination with `_WD_ExecuteScript` looks good.
- Should other coding in the related functions be updated to match this style?
- IMO, when you refer to `frame`, this implies both frames and iframes, so I don't see the need to list both in function header, array header, etc.

mlipok: Please take a look on:
[$sArrayHeader](https://github.com/Danp2/au3WebDriver/pull/434/commits/2b376e02e1cc572187d72f0ee0dca9509bfc40ce)
[header $sFrameAttributes](https://github.com/Danp2/au3WebDriver/pull/434/commits/9b1ef7a24d889ea090737f39fa045c0a1eada5ba)
[$sJavaScript...FrameAttributes...FrameElementID](https://github.com/Danp2/au3WebDriver/pull/434/commits/9ef8c2967a1ab3d69913db685231f998fa51b76f)

mlipok: anything more here ?

Danp2: Not AFAICS. I'm depending on you to perform testing on this since it's your baby.

P.S. Don't forget to remove your debugging code 😉

mlipok: > Not AFAICS. I'm depending on you to perform testing on this since it's your baby.

Just in any case, in future let me know when you giving a green light to clenaup and last review before merge.

> P.S. Don't forget to remove your debugging code 😉


thanks for remider.

Danp2:
@mlipok Apologies if we've discussed this previously, but can you explain the need for the `$iDebugLevel` parameter in __WD_FrameList_Internal? From what I can see --

- `_WD_FrameList` properly saves and restored the debug level.
- `__WD_FrameList_Internal` tries to "restore" the debug level before exiting, but hasn't changed it earlier in the function

Why does the debug level need to be tracked / modified in `__WD_FrameList_Internal`? 😕

mlipok:
> * `_WD_FrameList` properly saves and restored the debug level.

You mean this part:
https://github.com/Danp2/au3WebDriver/blob/bf98ff57ad5ab7afe4bd31ed0f93fdca96bff7b2/wd_helper.au3#L772-L778


> * `__WD_FrameList_Internal` tries to "restore" the debug level before exiting, but hasn't changed it earlier in the function
> 
> Why does the debug level need to be tracked / modified in `__WD_FrameList_Internal`? 😕


Becuase I want to get log to the console
https://github.com/Danp2/au3WebDriver/blob/bf98ff57ad5ab7afe4bd31ed0f93fdca96bff7b2/wd_helper.au3#L940

But I want to get them only once (when exiting main document) thus all this processing is needed.
https://github.com/Danp2/au3WebDriver/blob/bf98ff57ad5ab7afe4bd31ed0f93fdca96bff7b2/wd_helper.au3#L935-L941

mlipok:
> But I want to get them only once (when exiting main document) thus all this processing is needed.

We can get all error messages, maybe it will be better.
But still we need to know what level of debuging was set by user.

Thus I can move maintaining `$_WD_DEBUG_Saved` to inner of `__WD_FrameList_Internal`

mlipok:
> > But I want to get them only once (when exiting main document) thus all this processing is needed.
> 
> We can get all error messages, maybe it will be better. But still we need to know what level of debuging was set by user.

ALL >> I mean one log per each `__WD_FrameList_Internal` calls.

Danp2:
> >   Why does the debug level need to be tracked / modified in __WD_FrameList_Internal? 😕

> Becuase I want to get log to the console

OIC

> Thus I can move maintaining $_WD_DEBUG_Saved to inner of __WD_FrameList_Internal

Yes... as long as it isn't required in _WD_FrameList. 

> I mean one log per each __WD_FrameList_Internal calls.

Is that what you want?

mlipok:
Working on

mlipok:
> > > Why does the debug level need to be tracked / modified in __WD_FrameList_Internal? 😕
> 
> > Becuase I want to get log to the console
> 
> OIC
> 

Changed concept. but still prevnting redundant logs (most of logs)

> > Thus I can move maintaining $_WD_DEBUG_Saved to inner of __WD_FrameList_Internal
> 
> Yes... as long as it isn't required in _WD_FrameList.

the current commits only allow one possible log:
https://github.com/Danp2/au3WebDriver/blob/aa03f901325d359289f42452a46b2f2815b6870c/wd_helper.au3#L769

Which can occurs onces per each _WD_FrameList() call.

> 
> > I mean one log per each __WD_FrameList_Internal calls.
> 
> Is that what you want?

I rethink my approach.
it will be good to have at least one/finall message per each frame

https://github.com/Danp2/au3WebDriver/blob/aa03f901325d359289f42452a46b2f2815b6870c/wd_helper.au3#L940


Please take a look on this: [$_WD_DEBUG_Saved usage moved to internall](https://github.com/Danp2/au3WebDriver/pull/434/commits/aa03f901325d359289f42452a46b2f2815b6870c)

Danp2:
> Please take a look on this: [$_WD_DEBUG_Saved usage moved to internall](https://github.com/Danp2/au3WebDriver/pull/434/commits/aa03f901325d359289f42452a46b2f2815b6870c)

I like the direction this is headed towards as the code is much easier to follow IMO. I do have questions about this --
https://github.com/Danp2/au3WebDriver/blob/aa03f901325d359289f42452a46b2f2815b6870c/wd_helper.au3#L850-L852
and

https://github.com/Danp2/au3WebDriver/blob/aa03f901325d359289f42452a46b2f2815b6870c/wd_helper.au3#L930-L931

I'm having difficulty wrapping my head around how this will work. Won't `$_WD_DEBUG` get reset when exiting a nested frame? If so, then I don't think that is what you want.

mlipok:
WIP.....

image

```
__WD_FrameList_Internal: Error occurred on "null" level after processing subframe "null/12"
__WD_FrameList_Internal: Error occurred on "null" level when trying to check attributes of subframe "null/13"
__WD_FrameList_Internal: Error occurred on "null" level when trying to check attributes of subframe "null/14"
__WD_FrameList_Internal: Error occurred on "null" level when trying to check attributes of subframe "null/15"
__WD_FrameList_Internal: Error occurred on "null" level when trying to check attributes of subframe "null/16"
__WD_FrameList_Internal: Error occurred on "null" level when trying to check attributes of subframe "null/17"
__WD_FrameList_Internal: Error occurred on "null" level when trying to check attributes of subframe "null/18"
__WD_FrameList_Internal: Error occurred on "null" level when trying to check attributes of subframe "null/19"
__WD_FrameList_Internal: Error occurred on "null" level when trying to check attributes of subframe "null/20"
__WD_FrameList_Internal: Error occurred on "null" level when trying to check attributes of subframe "null/21"
__WD_FrameList_Internal: Error occurred on "null" level when trying to check attributes of subframe "null/22"
__WD_FrameList_Internal: Error occurred on "null" level when trying to check attributes of subframe "null/23"
````

**Comment by mlipok** [comment]
> WIP.....

still WIP

**Comment by mlipok** [comment]
> Won't `$_WD_DEBUG` get reset when exiting a nested frame? If so, then I don't think that is what you want.

Check current chnages from this: [usage of $_WD_DEBUG_Saved / $sMessage](https://github.com/Danp2/au3WebDriver/pull/434/commits/1dd0eeba142ab58e9344e6b7e336c8e53856e0a9)

btw.
My previous comment is not WIP ;)

File: wd_helper.au3 [hunk 3]
```diff
@@ -750,9 +750,7 @@ EndFunc   ;==>_WD_FrameLeave
 ;                  $bReturnAsArray      - [optional] Return result as array? Default is True.
 ; Return values .: Success - 2D array (with 7 cols) or string ( delimited with | and @CRLF ) @extended contains information about frame count
 ;                  Failure - "" (empty string) and sets @error to one of the following values:
-;                  - $_WD_ERROR_GeneralError
-;                  - $_WD_ERROR_Exception
-;                  - $_WD_ERROR_RetValue
+;                  - $_WD_ERROR_?????????
```
mlipok, line 753: need to be rethinked/checked

wd_helper.au3
```diff
@@ -832,7 +829,7 @@ EndFunc   ;==>_WD_FrameList
 ;                  $bIsHidden           - information about visibility of frame - taken by WebDriver
 ; Return values .: Success - string
 ;                  Failure - "" (empty string) and sets @error to one of the following values:
-;                  - $_WD_ERROR_Exception
+;                  - $_WD_ERROR_?????????
```
mlipok, line 832: need to be rethinked/checked

wd_helper.au3
Danp2: Other functions document this in one of the following manners --

```autoit
; Return values .: Success - <varies by function>
;                  Failure - Response from web driver and sets @error returned from _WD_ExecuteScript()
```

```autoit
; Return values .: Success - Matching Element ID(s)
;                  Failure - @error set to $_WD_ERROR_NoMatch if there are no matches OR
;                            Response from _WD_ExecuteScript() and sets @error to value returned from _WD_ExecuteScript()
```

wd_helper.au3
```diff
@@ -798,140 +786,158 @@ Func _WD_FrameList($sSession, $bReturnAsArray = True)
...
 		; select desired DataType for the $vResult - usually string is option for testing and asking support
 		If $bReturnAsArray Then
 			$vResult = $a_Result
 		Else
 			$vResult = _ArrayToString($a_Result)
 			If @error Then
 				$iErr = $_WD_ERROR_RetValue
-				$sMessage &= ' ArrayToString conversion failed'
+				$sMessage &= 'ArrayToString conversion failed. '
 				$vResult = ''
```
Danp2, line 797: $sMessage should be empty at this point so no need to perform concatenation, correct?

wd_helper.au3
```diff
@@ -798,140 +786,158 @@ Func _WD_FrameList($sSession, $bReturnAsArray = True)
...
-							$sMessage = 'Error occurred on "' & $sLevel & '" level when trying to check visibility of frames #' & $iFrame
+						If $iErr Then
+							$sMessage = 'Error occurred on "' & $sLevel & '" level when trying to check visibility of subframe "' & $sLevel & '/' & $iFrame & '"'
+							ContinueLoop
 						Else
-							$vResult &= __WD_FrameList_Internal($sSession, $sLevel & '/' & $iFrame, $sFrameAttributes, $bIsHidden, $iDebugLevel)
+							$vResult &= __WD_FrameList_Internal($sSession, $sLevel & '/' & $iFrame, $sFrameAttributes, $bIsHidden)
 							$iErr = @error
-							If Not @error Then
+							If $_WD_DEBUG_Saved <> $_WD_DEBUG_Full Then $_WD_DEBUG = $_WD_DEBUG_None ; Prevent logging multiple errors from __WD_FrameList_Internal - can be reset when back from inner/recurent __WD_FrameList_Internal() call
```
Danp2, line 915: Can you explain why this line is needed? Shouldn't prior call to `__WD_FrameList_Internal` have left `$_WD_DEBUG` unchanged?

wd_helper.au3
mlipok: Yes. But I wanted to have the same `design pattern` within function.

wd_helper.au3
mlipok: Thanks.
Please take a look: 
[redundant prevention removed](https://github.com/Danp2/au3WebDriver/pull/434/commits/89d50ff8dcb44828cdc1ec251f3970f34d80d32e)

wd_helper.au3
Danp2: That explanation doesn't make sense. Why would you concatenate unnecessarily just because you can and it "looks nice"?

mlipok: Ok.
Please take a look: [$sMessage &=](https://github.com/Danp2/au3WebDriver/pull/434/commits/fd2530dc28f76d79aaeae1b7c326f92a0101431a)

wd_helper.au3
mlipok: Please take a look: [I hope the final version](https://github.com/Danp2/au3WebDriver/pull/434/commits/8cac3267a123d1fd4e5f99d88332a054a8e074b6)

wd_helper.au3
Danp2: > I hope the final version

Wishful thinking 🤣 

I see you eliminated the code converting the array back into a string. That will lead to incorrect results because you manipulated the array contents following the array creation, correct?

wd_helper.au3
Danp2: That's a **really** long list of functions. Is it really necessary to list each function that was used by `__WD_FrameList_Internal`? What other alternatives did you consider?

wd_helper.au3
mlipok: Please take a look on: [revert: getting string with recalculated locations](https://github.com/Danp2/au3WebDriver/pull/434/commits/b4d9ce5945fe14a3467546dab173a04b3880a9b1)

Danp2 (changes requested):
IMO, a function doesn't belong in the related section simply because it is called internally by the UDF being documented. And this information gets used to generate links in the help file, and it doesn't make sense to me to have links to many of these functions from _WD_FrameList. I think that we should evaluate if this is the proper method of documenting these error codes.

wd_helper.au3
Danp2: See my review below.

mlipok: Asap

mlipok:
> I think that we should evaluate if this is the proper method of documenting these error codes.

Please propose solution by adding new commit.

wd_helper.au3
mlipok: Cool

CHANGELOG.md
```diff
@@ -17,6 +17,10 @@ Go to [legend](#legend---types-of-changes) for further information about the typ
 - _WD_Option: Support for `DetailErrors` option
 
 ### Changed
+- _WD_FrameList
+	- Refactored for better performance
+	- Improved frame support
+	- Improved logging
```
Danp2, line 23: @mlipok Please review this for content / accuracy. Also need to know if you wanted `_WD_FrameList` to utilize the new detailed error reporting.

mlipok: > Also need to know if you wanted `_WD_FrameList` to utilize the new detailed error reporting.

Yes

> Please review this for content / accuracy.

will look today

mlipok: > Also need to know if you wanted `_WD_FrameList` to utilize the new detailed error reporting.

Please take a look: 
[$_WD_DetailedErrors support](https://github.com/Danp2/au3WebDriver/pull/434/commits/fe5c7507ef27b20f895b5028082145a7fc840326)

mlipok: > > Also need to know if you wanted `_WD_FrameList` to utilize the new detailed error reporting.
> 
> Yes
> 
> > Please review this for content / accuracy.
> 
> will look today

I'm fine with this changes.

Danp2: See [minor tweaks](https://github.com/Danp2/au3WebDriver/pull/434/commits/7ca53a1d3f6056e18d25849be55ffef8c1926fb1)

Danp2: > > Also need to know if you wanted `_WD_FrameList` to utilize the new detailed error reporting.
> 
> Yes

There are additional places in this function where the value of $iErr is modified. I think some of them should also be updated. Please review to see if you agree.

mlipok: > See [minor tweaks](https://github.com/Danp2/au3WebDriver/pull/434/commits/7ca53a1d3f6056e18d25849be55ffef8c1926fb1)

It looks fine

Issue #453, opened 2023-04-19
IEDriverServer.exe support


mlipok:
### Is your feature request related to a problem? Please describe

no

### Describe the solution you'd like

we should add support for IEDriverServer.exe in UDF's

### Describe alternatives you've considered

none

### Additional context

https://www.autoitscript.com/forum/topic/210059-website-giving-error-using-script-but-fine-manually
https://www.autoitscript.com/forum/topic/205553-webdriver-udf-help-support-iii/?do=findComment&comment=1500599

https://www.selenium.dev/documentation/ie_driver_server/#required-configuration
https://learn.microsoft.com/en-us/answers/questions/761761/how-to-disable-ie-mode-force-mode-in-iedriverserve

currently working example is available here:
https://github.com/mlipok/Au3WebDriver-testing

Danp2:
My general thoughts / questions --

- Demand for this appears low.
- Existing functionality is focused on browser support rather than driver support. Are you wanting to support Edge in IE mode or something else?
- Would any of the existing functions require modification? If yes, which ones?

mlipok:
> > * Demand for this appears low.
> 
> comparable to OPERA ?
> 
> > * Existing functionality is focused on browser support rather than driver support. Are you wanting to support Edge in IE mode or something else?
> 
> EDGE support it so we should be able to use EDGE with IE Mode

any comment on this ?


> > * Would any of the existing functions require modification? If yes, which ones?
> 
> I will look at.

It should do almost the same as in: https://github.com/Danp2/au3WebDriver/pull/213

mlipok:
btw. I already did some tweaks and have updated 
```autoit
Global Const $_WD_SupportedBrowsers[][$_WD_BROWSER__COUNTER] = _
		[ _
		["chrome", "chrome.exe", "chromedriver.exe", False, "goog:chromeOptions", "'https://chromedriver.storage.googleapis.com/LATEST_RELEASE_' & StringLeft($sBrowserVersion, StringInStr($sBrowserVersion, '.') - 1)", "", '"https://chromedriver.storage.googleapis.com/" & $sDriverLatest & "/chromedriver_win32.zip"'], _
[...]

PR #454, opened 2023-04-20
IEDriverServer.exe "iemode" support

mlipok:
https://github.com/Danp2/au3WebDriver/issues/453

### Proposed changes

support for IEDriverServer.exe "iemode" 

### Checklist

### Types of changes

- [ ] Other (please describe)

no support for IEDriverServer.exe "iemode"

### What is the new behavior?

added support for IEDriverServer.exe "iemode"

### Additional context

https://github.com/Danp2/au3WebDriver/issues/453

### System under test

Browser MSEdge in "iemode"


mlipok:
https://github.com/Danp2/au3WebDriver/issues/453#issuecomment-1516330864

> Did you have success when testing this with wd_demo? For me, the browser launches but no other commands appear to work. Tried both 32/64bit versions of IEDriverServer.

Please let continue here.

you mean this:
```
+ wd_demo.au3: Running: DemoNavigation
__WD_Post ==> Send / Recv error [6] : HTTP status = 0
_WD_Navigate ==> Send / Recv error [6] : Parameters:   URL=https://google.com
_WD_LoadWait ==> Webdriver Exception [10 / 0] : Parameters:    Delay=Default    Timeout=Default    Element=Default    DesiredState=complete    : ReadyState= 
_WD_WaitElement ==> Send / Recv error [6] : Parameters:   Strategy=xpath   Selector=//body/div[1][@aria-hidden="true"]   Delay=0   Timeout=60000   Options=8
```

mlipok:
Using this: [-log-level=INFO](https://github.com/Danp2/au3WebDriver/pull/454/commits/108ca3eac1b72b872acc52d1dcfc309bce50e14c)

I get such results:
```
I 2023-04-20 17:04:50:488 C:\projects\seleniumhq\selenium\cpp\webdriver-server\server.cc(88) Starting WebDriver server on port: '5555' on host: '127.0.0.1'
I 2023-04-20 17:04:50:488 C:\projects\seleniumhq\selenium\cpp\iedriver\IEServer.cpp(33) Driver version: 4.8.1.0 (32-bit)Started InternetExplorerDriver server (32-bit)
4.8.1.0
Listening on port 5555
Bound to network adapter with IP address 127.0.0.1
Log level is set to INFO
Only local connections are allowed
W 2023-04-20 17:04:51:741 C:\projects\seleniumhq\selenium\cpp\iedriver\CommandHandlers\NewSessionCommandHandler.cpp(105) Invalid capability setting: timeouts is type null instead of object. Default value will be used: null

I 2023-04-20 17:04:52:586 C:\projects\seleniumhq\selenium\cpp\iedriver\IECommandExecutor.cpp(1604) Persistent hovering set to: 1
I 2023-04-20 17:04:52:602 C:\projects\seleniumhq\selenium\cpp\iedriver\Browser.cpp(239) No child frame focus. Focus is on top-level frame
W 2023-04-20 17:04:52:724 C:\projects\seleniumhq\selenium\cpp\iedriver\Browser.cpp(67) A new browser process was requested. This means a Protected Mode boundary has been crossed, and that future commands to the current browser instance will fail. The driver will attempt to reconnect to the newly created browser object, but there is no guarantee it will work.
```

Danp2 (changes requested):
The port hasn't been set yet, so this will always be 0. This line needs to be moved below the line `_WD_Option('Port', 5555)` for this to work.

mlipok:
This 
https://www.edureka.co/community/1690/problem-using-internetexplorerdriver-selenium-webdriver

help me fixing the issue with navigating.

mlipok:
> [@mlipok ](/mlipok)
> ` ['Port' setting moved](/Danp2/au3WebDriver/pull/454/commits/8ca92d2ebc42c2170d64613f4d1af946ec56e076)`


Please take a look: 
['Port' setting moved](https://github.com/Danp2/au3WebDriver/pull/454/commits/8ca92d2ebc42c2170d64613f4d1af946ec56e076)

mlipok:
> This https://www.edureka.co/community/1690/problem-using-internetexplorerdriver-selenium-webdriver
> 
> help me fixing the issue with navigating.

my results are:
```
+ wd_demo.au3: Running: DemoNavigation
wd_demo.au3: (339) : URL=https://consent.google.com/ml?continue=https://www.google.com/&gl=PL&m=0&pc=shp&uxe=none&hl=pl&src=1
wd_demo.au3: (344) : URL=https://www.yahoo.com/
wd_demo.au3: (347) : URL=https://www.bing.com/?toWww=1&redig=734236F6925643DC9EF5801740F5793B
wd_demo.au3: (350) : URL=https://consent.google.com/ml?continue=https://www.google.com/&gl=PL&m=0&pc=shp&uxe=none&hl=pl&src=1
wd_demo.au3: (353) : URL=https://www.yahoo.com/
+ wd_demo.au3: Finished: DemoNavigation
```

mlipok:
it could be sets in registry:
https://learn.microsoft.com/en-us/dynamics365/unified-service-desk/admin/internet-explorer-settings-bpa?view=dynamics-usd-4.2

Danp2:
> A new browser process was requested. This means a Protected Mode boundary has been crossed, and that future commands to the current browser instance will fail. The driver will attempt to reconnect to the newly created browser object, but there is no guarantee it will work.

I believe this is due to this setting, which is part of "se:ieOptions" --

`"initialBrowserUrl" : "http://localhost:5555/"`

Have you tried setting this to something that wouldn't be considered "local" prior to launching the session? IMO, adding sites to your trusted zone isn't a viable solution.

wd_demo.au3
```diff
@@ -1263,3 +1264,47 @@ Func SetupOpera($bHeadless)
...
+			'        }' & @CRLF & _
+			'    }' & @CRLF & _
+			'}' & @CRLF & _
+			''
+	ConsoleWrite("! $sCapabilities (by hand) = " & @CRLF & $sCapabilities & @CRLF)
+#ce
+
+	_WD_CapabilitiesDefine($_WD_KEYS__SPECIFICVENDOR_PRIMITIVE, "ie.edgepath")
+	_WD_CapabilitiesDefine($_WD_KEYS__SPECIFICVENDOR_PRIMITIVE, "ie.edgechromium")
+	_WD_CapabilitiesDefine($_WD_KEYS__SPECIFICVENDOR_PRIMITIVE, "ignoreProtectedModeSettings")
```
Danp2, line 1297: If these are required for IEMode to work correctly, then they should be included by default.

wd_demo.au3
```diff
@@ -1263,3 +1264,47 @@ Func SetupOpera($bHeadless)
...
+
+	_WD_CapabilitiesStartup()
+	_WD_CapabilitiesAdd('alwaysMatch', 'msedgeie')
+	_WD_CapabilitiesAdd('w3c', True)
+	_WD_CapabilitiesAdd("ie.edgepath", "C:\Program Files (x86)\Microsoft\Edge\Application\msedge.exe")
```
Danp2, line 1302: Use `_WD_GetBrowserPath` instead of hardcoding the path.

mlipok:
> I believe this is due to this setting, which is part of "se:ieOptions" --
> 
> `"initialBrowserUrl" : "http://localhost:5555/"`
> 
> Have you tried setting this to something that wouldn't be considered "local" prior to launching the session? IMO, adding sites to your trusted zone isn't a viable solution.

so I used:
```auotit
	_WD_CapabilitiesDefine($_WD_KEYS__SPECIFICVENDOR_PRIMITIVE, "initialBrowserUrl")
	_WD_CapabilitiesAdd("initialBrowserUrl", "about:blank")
```
and I get:
```
I 2023-04-20 17:56:04:182 C:\projects\seleniumhq\selenium\cpp\webdriver-server\server.cc(88) Starting WebDriver server on port: '5555' on host: '127.0.0.1'
I 2023-04-20 17:56:04:183 C:\projects\seleniumhq\selenium\cpp\iedriver\IEServer.cpp(33) Driver version: 4.8.1.0 (32-bit)Started InternetExplorerDriver server (32-bit)
4.8.1.0
Listening on port 5555
Bound to network adapter with IP address 127.0.0.1
Log level is set to INFO
Only local connections are allowed
W 2023-04-20 17:56:05:135 C:\projects\seleniumhq\selenium\cpp\iedriver\CommandHandlers\NewSessionCommandHandler.cpp(105) Invalid capability setting: timeouts is type null instead of object. Default value will be used: null
```

but:

```
> wd_demo.au3: _WD_CreateSession
__WD_Post ==> Send / Recv error [6] : HTTP status = 0
_WD_CreateSession ==> Webdriver Exception [10]
````

**Comment by mlipok** [comment]
If I revert back the ZONES settings and after using:
```autoit
_WD_CapabilitiesAdd("initialBrowserUrl", "https://google.com")
```

all looks fine, take a look here:  [initialBrowserUrl](https://github.com/Danp2/au3WebDriver/pull/454/commits/9f6d24e16bc0acfbce4252cbf300616984890b98)

wd_demo.au3
mlipok: Take a look here: 
[_WD_GetBrowserPath("msedge")](https://github.com/Danp2/au3WebDriver/pull/454/commits/6b84292ac8d6b89c02b45d98084771afa090a01f)

wd_demo.au3
Danp2: IMO, they should already be defined so that the end user should be forced to add these via `_WD_CapabilitiesDefine`.

wd_demo.au3
```diff
@@ -1263,3 +1264,50 @@ Func SetupOpera($bHeadless)
...
 EndFunc   ;==>SetupOpera
+
+Func SetupEdgeIEMode() ; this is for MS Edge IE Mode
+	#REMARK In order to properly use MS Edge In IE Mode you need To Do some settings - Set the "Internet, Local intranet, Trusted sites, Restricted sites" to protected mode, enabling or disabling it is not a matter
+	; https://www.edureka.co/community/1690/problem-using-internetexplorerdriver-selenium-webdriver
+	Local $sTimeStamp = @YEAR & '-' & @MON & '-' & @MDAY & '_' & @HOUR & @MIN & @SEC
+	; https://www.selenium.dev/documentation/ie_driver_server/#required-configuration
+	_WD_Option('Driver', 'IEDriverServer.exe') ;
+	_WD_Option('Port', 5555)
+	_WD_Option('DriverParams', '--verbose --log-path="' & @ScriptDir & '\' & $sTimeStamp & ' EdgeIEMode.log" -log-level=INFO' & " --port=" & $_WD_PORT & " -host=127.0.0.1" )
```
Danp2, line 1275: - According to [here](https://www.selenium.dev/documentation/ie_driver_server/), this should be `-log-file` instead of `--log-path`. 
- It doesn't look like `--verbose` is valid either.
- Need to check `-port` vs `--port`.

mlipok: > * According to [here](https://www.selenium.dev/documentation/ie_driver_server/), this should be `-log-file` instead of `--log-path`.
> * It doesn't look like `--verbose` is valid either.

already tested:
any change you pointed me to, will effect in not creating log files.

Danp2: Hmmm... suggest that you check again because I just tried it with the recommended switch and the log file was successfully created.

mlipok: Check this commit: [DriverParams usage](https://github.com/Danp2/au3WebDriver/pull/454/commits/36486fa731192afe31f6fe7f0d66549c521b5462)

wd_demo.au3
mlipok: after removig:
```
_WD_CapabilitiesDefine($_WD_KEYS__SPECIFICVENDOR_PRIMITIVE, "ie.edgepath")
```

all was fine, but after removing:
```autoit
_WD_CapabilitiesDefine($_WD_KEYS__SPECIFICVENDOR_PRIMITIVE, "ie.edgechromium")
```
I starts getting issues:
image

mlipok: using:
```autoit
	_WD_CapabilitiesAdd('alwaysMatch', 'msedgeie')
	_WD_CapabilitiesAdd('w3c', True)
	_WD_CapabilitiesAdd("ie.edgechromium", True)
;~ 	_WD_CapabilitiesAdd("ignoreProtectedModeSettings", True)
	_WD_CapabilitiesAdd("initialBrowserUrl", "https://google.com")
	_WD_CapabilitiesAdd('excludeSwitches', 'enable-automation')
```
I get:
```
! _WD_LastHTTPResponse = {
	"value" : 
	{
		"error" : "session not created",
		"message" : "Unexpected error launching Internet Explorer. Protected Mode settings are not the same for all zones. Enable Protected Mode must be set to the same value (enabled or disabled) for all zones.",
		"stacktrace" : ""
	}
}
```

mlipok: so finally: [ie.edgechromium - not needed](https://github.com/Danp2/au3WebDriver/pull/454/commits/e0e25c25fa5b459fe1b6d28c8c3a86061e99be2e)

wd_demo.au3
```diff
@@ -1266,6 +1266,8 @@ Func SetupOpera($bHeadless)
 EndFunc   ;==>SetupOpera
 
 Func SetupEdgeIEMode() ; this is for MS Edge IE Mode
+	#REMARK In order to properly use MS Edge In IE Mode you need To Do some settings - Set the "Internet, Local intranet, Trusted sites, Restricted sites" to protected mode, enabling or disabling it is not a matter
+	; https://www.edureka.co/community/1690/problem-using-internetexplorerdriver-selenium-webdriver
```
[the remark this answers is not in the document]
mlipok: [REMARK removed](https://github.com/Danp2/au3WebDriver/pull/454/commits/54c1c0d9a6e11c11c967acf6a27a3383f2f9e4b5)

wd_demo.au3
Danp2: Ok, but that doesn't satisfy my objection

> they should already be defined so that the end user isn't forced to add these via _WD_CapabilitiesDefine.

Do you still not understand what I am wanting here?

mlipok: Do you mean that instead using:
```autoit
	_WD_CapabilitiesDefine($_WD_KEYS__SPECIFICVENDOR_PRIMITIVE, "ie.edgepath")
	_WD_CapabilitiesDefine($_WD_KEYS__SPECIFICVENDOR_PRIMITIVE, "ie.edgechromium")
	_WD_CapabilitiesDefine($_WD_KEYS__SPECIFICVENDOR_PRIMITIVE, "ignoreProtectedModeSettings")
	_WD_CapabilitiesDefine($_WD_KEYS__SPECIFICVENDOR_PRIMITIVE, "initialBrowserUrl")
```
I shoud change wd_capabilities.au3
```autoit
Global $_WD_KEYS__SPECIFICVENDOR_PRIMITIVE = _
		'\A(binary|debuggerAddress|detach|minidumpPath|w3c|ie.edgepath|ie.edgechromium|ignoreProtectedModeSettings|initialBrowserUrl)\Z'
```

mlipok: So take a look on:
[demo: _WD_CapabilitiesDefine - removed](https://github.com/Danp2/au3WebDriver/pull/454/commits/071067ff65b303693a0eaa5d376a54c7e0af4c74)

and
[Global $_WD_KEYS__SPECIFICVENDOR_PRIMITIVE](https://github.com/Danp2/au3WebDriver/pull/454/commits/d793353acafbfcf3c105a814cf8fedce2b0c9b6a)

mlipok:
> The port hasn't been set yet, so this will always be 0. This line needs to be moved below the line `_WD_Option('Port', 5555)` for this to work.

already fixed:
https://github.com/Danp2/au3WebDriver/blob/d793353acafbfcf3c105a814cf8fedce2b0c9b6a/wd_demo.au3#L1272-L1273

wd_demo.au3
```diff
@@ -1263,3 +1264,41 @@ Func SetupOpera($bHeadless)
...
 	Return $sCapabilities
 EndFunc   ;==>SetupOpera
+
+Func SetupEdgeIEMode() ; this is for MS Edge IE Mode
+	; https://www.selenium.dev/documentation/ie_driver_server/#required-configuration
+	Local $sTimeStamp = @YEAR & '-' & @MON & '-' & @MDAY & '_' & @HOUR & @MIN & @SEC
+	_WD_Option('Driver', 'IEDriverServer.exe') ;
+	_WD_Option('Port', 5555)
+	_WD_Option('DriverParams', '-log-file="' & @ScriptDir & '\log\' & $sTimeStamp & '_WebDriver_EdgeIEMode.log" -log-level=INFO' & " -port=" & $_WD_PORT & " -host=127.0.0.1" )
+#cs
```
mlipok, line 1274: should we remove the commented section ?

Danp2: I would keep the first line and remove the formatted version. That way it matches the other "setup" routines.

mlipok: what abou the `ie.edgepath` usage should we revert them back ?

Danp2: I would leave it there to show that it's an option that could be used if needed.

mlipok: Done
[SetupEdgeIEMode - cleanup](https://github.com/Danp2/au3WebDriver/pull/454/commits/defe8af3f738cdaa5398230325cc80d0e3224a92)

wd_demo.au3
```diff
@@ -1263,3 +1264,25 @@ Func SetupOpera($bHeadless)
 	Local $sCapabilities = _WD_CapabilitiesGet()
 	Return $sCapabilities
 EndFunc   ;==>SetupOpera
+
+Func SetupEdgeIEMode() ; this is for MS Edge IE Mode
+	; https://www.selenium.dev/documentation/ie_driver_server/#required-configuration
+	Local $sTimeStamp = @YEAR & '-' & @MON & '-' & @MDAY & '_' & @HOUR & @MIN & @SEC
+	_WD_Option('Driver', 'IEDriverServer.exe') ;
+	_WD_Option('Port', 5555)
```
Danp2, line 1272: I've noticed something odd in the logs after running the IE driver more than once --

```
    _WD_GetFreePort ==> Not found [13]
    _WD_IsLatestRelease ==> Success [0] : True
    _WD_Startup: OS:	WIN_11 X64 22621 
    _WD_Startup: AutoIt:	3.3.16.1
    _WD_Startup: Webdriver UDF:	0.13.0 (Up to date)
    _WD_Startup: WinHTTP:	1.6.4.2
    _WD_Startup: Driver:	IEDriverServer.exe (32 Bit)
    _WD_Startup: Params:	-log-file="c:\Users\danpo\OneDrive\Documents\GitHub\WebDriver\log\2023-04-20_133006_WebDriver_EdgeIEMode.log" -log-level=INFO -port=5555 -host=127.0.0.1
    _WD_Startup: Port:	5555 (Unavailable)
```

This doesn't appear to affect the driver's functionality. Unsure if this is an issue with IEDriverServer or something is wrong elsewhere.

Danp2: This is the output from netstat --

```
netstat -aon | find "5555"
  TCP    127.0.0.1:5555         127.0.0.1:60325        TIME_WAIT       0
```

It does appear to timeout after a while, so I guess it's NBD.

mlipok: image

Danp2: > you mean that this port is blocked ?

Not blocked, but it is detected as "in use" by `_WD_GetFreePort`. 

> maybe IE is still in background ?

Nope, and we've both observed that the port is eventually released after the timeout occurs. AFAIK, this doesn't occur with the other webdrivers. However, I may be able to modify `_WD_GetFreePort` so that it ignores ports in the TIME_WAIT state.

mlipok:
```
C:\Windows\System32>netstat -aon | find "5555"
  TCP    127.0.0.1:5555         0.0.0.0:0              LISTENING       3000
  TCP    127.0.0.1:5555         127.0.0.1:52102        ESTABLISHED     3000
  TCP    127.0.0.1:52102        127.0.0.1:5555         ESTABLISHED     26652
```
PID 3000 => IEDriverServer.exe
PID 26652 => AutoIt3_x64.exe

mlipok: Please check this:
```auotit
_WD_Option('Port', _WD_GetFreePort(5555, 5600))
```

mlipok: Take a look on:
[_WD_GetFreePort usage](https://github.com/Danp2/au3WebDriver/pull/454/commits/9e02585fdc28493a4c954eb8c1b347ae8d76ae91)

wd_demo.au3
```diff
@@ -1269,7 +1269,7 @@ Func SetupEdgeIEMode() ; this is for MS Edge IE Mode
 	; https://www.selenium.dev/documentation/ie_driver_server/#required-configuration
 	Local $sTimeStamp = @YEAR & '-' & @MON & '-' & @MDAY & '_' & @HOUR & @MIN & @SEC
 	_WD_Option('Driver', 'IEDriverServer.exe') ;
-	_WD_Option('Port', 5555)
+	_WD_Option('Port', _WD_GetFreePort(5555, 5600))
```
mlipok, line 1272: should we add 
```autoit
_WD_Option('Port', _WD_GetFreePort(5555, 5600))
```
also for other browsers ?

Danp2: I don't think so. Technically, it isn't even needed for `iemode` because everything still runs fine, albeit with some minor logging glitches.

mlipok: IMO it is possible that `iemod` on some slower PC could hit the problem very often and will be unable to start driver connection.

Danp2: It has been my experience the IEDriverServer would load and run even when the target port is in  TIME_WAIT state. I can immediately run IEDriverServer after it has exited a prior run.

Does that match your experience?

mlipok: Please try with this modified:

```
Func _WD_GetFreePort($iMinPort = Default, $iMaxPort = Default)
	Local Const $sFuncName = "_WD_GetFreePort"
	Local $iResult = 0, $iErr = $_WD_ERROR_Success

	If $iMaxPort = Default Then $iMaxPort = ($iMinPort = Default) ? 65000 : $iMinPort
	If $iMinPort = Default Then $iMinPort = 64000
	Local $aPorts = __WinAPI_GetTcpTable()
	If @error Then
	Else
		For $iPort = $iMinPort To $iMaxPort
			_ArraySearch($aPorts, $iPort, Default, Default, Default, Default, Default, 3)
			If @error = 6 Then ; not found in list of used port
				$iResult = $iPort
				ExitLoop
			EndIf
		Next
		If $iPort > $iMaxPort Then $iErr = $_WD_ERROR_NotFound
	EndIf

	Return SetError(__WD_Error($sFuncName, $iErr, $iResult), 0, $iResult)
EndFunc   ;==>_WD_GetFreePort
```

and:
```autoit
	_WD_Option('Driver', 'IEDriverServer.exe') ;
	_WD_Option('Port', _WD_GetFreePort(5555))
```

note:
I think that you hit an error from: `__WinAPI_GetTcpTable()`
IMO we should get two different errors from: _WD_GetFreePort 
$_WD_ERROR_GeneralError 
or 
$_WD_ERROR_NotFound

mlipok: > It has been my experience the IEDriverServer would load and run even when the target port is in TIME_WAIT state. I can immediately run IEDriverServer after it has exited a prior run.


I have used:
```autoit
	Local $aPorts = __WinAPI_GetTcpTable()
	_ArrayDisplay($aPorts)
```
and tested
```
C:\Windows\System32>netstat -aon | find "5555"
  TCP    127.0.0.1:5555         127.0.0.1:53183        TIME_WAIT       0
  TCP    127.0.0.1:5555         127.0.0.1:53252        TIME_WAIT       0
```
and `netstat` shows what you see above and `_ArrayDisplay` is showing this port as used.
image



> Does that match your experience?

No as I had no issue with: `_WD_Startup: Port: 5555 (Unavailable)`
which as I stated comes from: `__WinAPI_GetTcpTable()`

Danp2: > No as I had no issue with: _WD_Startup: Port: 5555 (Unavailable)

Run wd_demo, run multiple passes of `DemoNavigation`, then check the Scite output panel. Obviously, you would need to revert to `_WD_Option('Port', 5555)` for the test to be valid.

mlipok: > It has been my experience the IEDriverServer would load and run even when the target port is in TIME_WAIT state. I can immediately run IEDriverServer after it has exited a prior run.
> 
> Does that match your experience?

yes. I confirm

Danp2: > 	_WD_Option('Port', _WD_GetFreePort(5555))

There's no point in calling `_WD_GetFreePort` here (when a single port is provided) as it is automatically called in `_WD_Startup`. Doing it this way will assign 0 as the port number if the port is in use.

mlipok: oops 
```autoit
_WD_Option('Port', _WD_GetFreePort(5555, 5600))
```

Danp2: > and netstat shows what you see above and _ArrayDisplay is showing this port as used.

Right... which would explain why I received the "Unavailable" message because `_WD_GetFreePort` detected that the port was open. I agree with adding the additional error code. I'll submit a separate PR for this and other possible _WD_GetFreePort enhancements.

mlipok: or even better
```autoit
_WD_Option('Port', _WD_GetFreePort())
```

Danp2: > _WD_Option('Port', _WD_GetFreePort(5555, 5600))

I'm thinking that these should be separated into two statements, so that you could handle any errors. Otherwise, the error gets ignored and the port will equal 0.

Something like this --

```
Local $iPort = _WD_GetFreePort(5555, 5600))
If @error Then $iPort = 5555
_WD_Option('Port', $iPort)
```

mlipok: Here is better idea:

[Execute browser setup - error handling](https://github.com/Danp2/au3WebDriver/pull/454/commits/aa592a1f12ac8e4568a5e99a868684a00c81fbb2)

[SetupEdgeIEMode + _WD_GetFreePort](https://github.com/Danp2/au3WebDriver/pull/454/commits/aa458ff0d3725f2fd44750eff0a0d2a3f6c16cc6)

test it with:
```autoit
Local $iPort = _WD_GetFreePort(5555, 5555)
```

Danp2: FWIW, I like my way better because it doesn't abort due to an issue with `__WinAPI_GetTcpTable`. It also should be easier for the user to determine what went wrong, whereas your version simply exits without any type of explanation AFAICS.

mlipok: > because it doesn't abort due to an issue with `__WinAPI_GetTcpTable`.

if the issue occurs then port will be 0, we can check if the error is general or not found and in case generall trying to set it manually.

mlipok: > whereas your version simply exits without any type of explanation AFAICS.

`_WD_GetFreePort` should be modified to provide `$sMessage`

mlipok: We can also force to return `$iMinPort` on `$_WD_ERROR_NotFound`

mlipok:
Tested with DemoFrames and https://github.com/Danp2/au3WebDriver/pull/456

wd_demo.au3
Danp2: > _WD_GetFreePort should be modified to provide $sMessage

It looks like the current call to `__WD_Error` is currently missing that parameter. 

> if the issue occurs then port will be 0, we can check if the error is general or not found and in case generall trying to set it manually.

This sounds like what I suggested, but with a check for which error occurred.

> In case $_WD_ERROR_NotFound we can also force to return $iMinPort instead 0

Yes, that is certainly an option.

Danp2: See #457

mlipok: > See #457

Saw.
It will be ok all together.

mlipok:
So as for me all is done.

PR #484, opened 2023-07-20
chrome >> chrome_legacy - automatically supported

mlipok:
### Proposed changes

chrome >> chrome_legacy - should be automatically supported user should not bother about this

### Checklist

### Types of changes

- [ ] Other (please describe)

you need to check which version of Chrome browser you have and specify this by choosing **chrome** or **chrome_legacy** 

### What is the new behavior?

chrome >> chrome_legacy - automatically done

### Additional context

https://www.autoitscript.com/forum/topic/208640-webdriver-udf-help-support-iv/?do=findComment&comment=1521299

https://www.autoitscript.com/forum/topic/208640-webdriver-udf-help-support-iv/?do=findComment&comment=1521377

### System under test
Chrome Browser v114... v115

### REMARK
consider to check for `$_WD_SupportedBrowsers` if there is a need to add accordingly some additionall checking

### DIFF
--- a/wd_helper.au3
+++ b/wd_helper.au3
@@ -2081,7 +2081,7 @@ Func _WD_UpdateDriver($sBrowser, $sInstallDir = Default, $bFlag64 = Default, $bF
 		If $iErr = $_WD_ERROR_Success Then
 			Local $iIndex = @extended
 			If $sBrowser = "chrome" Then
-				Local $i_Check = (_VersionCompare("115.0.0.0", $sBrowserVersion)
+				Local $i_Check = _VersionCompare("115.0.0.0", $sBrowserVersion)
 				If Not @error And $i_Check = 1 Then $sBrowser = "chrome_legacy"
 			EndIf
 			; Match exe file name in list of supported browsers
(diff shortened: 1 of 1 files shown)


mlipok:
> The `chrome_legacy` entry will be removed in the near future once everyone has had the opportunity to upgrade to newer versions of Chrome.

Why in near future ?
I would like to prospose to leave this functionality as long as chromedriver.exe download links are available/active.

Danp2:
> So as I understand your statement user should check it by his own route

There are several way to tackle this problem --

- Read the target browser from an INI file, where the end user could easily switch between `chrome` and `chrome_legacy`
- Perform the check (similar to what you posted above) at the beginning of the script to determine `chrome` vs `chrome_legacy`. Then use this stored value throughout the rest of the script.

> Why in near future ?
I would like to prospose to leave this functionality as long as chromedriver.exe download links are available/active.

For multiple reasons --

- As indicated in the change log, this was a solution that was temporarily added to support both old and new download locations of Chromedriver.
- IMO, it is unrealistic to expect the UDF to continue to support an outdated browser.
- Anyone needing this "legacy support" can maintain their own copy of the UDF

mlipok:
Ok.
I have one conception to do, propose, and we will be able to close this PR in max next 2 days.

mlipok:
As for now it can be closed.

PR #517, opened 2024-04-25
_WD_UpdateDriver - desired BrowserVersion

mlipok:
### Proposed changes

User should be able to provide specific/desired browser version to download/update desired driver

### Checklist

### Types of changes

- [X] Feature (change which adds functionality)
- [ ] Other (please describe)

computer station should be connected to internet to be able to get driver related to current computer station

### What is the new behavior?

you can check browser version computer only with INTRANET (not connected to internet) store them for example to INI file, 
and downlad the desired driver on different computer

none

### Additional context
https://www.autoitscript.com/forum/topic/210991-webdriver-_wd_updatedriverfirefox-dont-work-on-intranet

https://www.autoitscript.com/forum/topic/210991-webdriver-_wd_updatedriverfirefox-dont-work-on-intranet/?do=findComment&comment=1525803

### System under test
not related

### DIFF
--- a/wd_helper.au3
+++ b/wd_helper.au3
@@ -2035,12 +2035,13 @@ EndFunc   ;==>_WD_IsLatestRelease
 ; #FUNCTION# ====================================================================================================================
 ; Name ..........: _WD_UpdateDriver
 ; Description ...: Replace web driver with newer version, if available.
-; Syntax ........: _WD_UpdateDriver($sBrowser[, $sInstallDir = Default[, $bFlag64 = Default[, $bForce = Default[, $bDowngrade = Default]]]])
-; Parameters ....: $sBrowser    - Browser name or full path to browser executable
-;                  $sInstallDir - [optional] Install directory. Default is @ScriptDir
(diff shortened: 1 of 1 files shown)


Danp2:
@mlipok This looks good to me. One thing to consider is what happens if an invalid browser version is supplied by the user? Does some form of validity check need to be performed on this parameter?

Sven-Seyfert:
Hi to both of you 👋 ,

> One thing to consider is what happens if an invalid browser version is supplied by the user? Does some form of validity check need to be performed on this parameter?

My honest answer is no. Why should we/you provide a mechanism to validate such specific user input? In case you define a specific browser version, I am quite sure that the user have to check the correct version on themself. Of course I understand your aspect of providing almost the best and most comfortable way for the user, but I didn't see such "helping hand" in Selenium, WebdriverIO or Playwright.

It could also be very tricky and error prone to add a good validation, because of the provided browser versions of the provider webpages etc. And: it would increase the complexity of the code which is in **this case** not necessary.

But: This is only my opinion and 2 Cents for that 😇 .

Best regards
Sven

@Danp2 
@mlipok

PR #524, opened 2024-09-10
_WD_JsonActionKey: Don't encode Unicode characters

Danp2:
### Proposed changes

Don't encode the backslash character when unicode character is used with _WD_JsonActionKey

### Checklist

### Types of changes

- [ ] Other (please describe)

Unicode characters are encoded so that they begin with a double backslash

### What is the new behavior?

Check for unicode character and prevent double backslash

### Additional context

https://www.autoitscript.com/forum/topic/212262-webdriver-udf-_wd_elementactionex-reports-errors-when-trying-to-use-command-modifierclick/

### System under test

- OS: [e.g. Windows 10]
- OS Arch.: [e.g. X64]
- Browser [e.g. firefox]

Signed-off-by: Dan Pollak <danpollak2@gmail.com>

### DIFF
--- a/wd_helper.au3
+++ b/wd_helper.au3
@@ -3221,6 +3221,12 @@ Func _WD_JsonActionKey($sType, $sKey, $iSuffix = Default)
 	Json_Put($vData, '.actions[0].type', $sType)
 	Json_Put($vData, '.actions[0].value', $sKey)
 	Local $sJSON = Json_Encode($vData)
+
+	; Don't encode backslash of Unicode character
+	If StringLeft($sKey,2) = '\u' Then
+		$sJSON = StringReplace($sJSON, "\\u", "\u")
+	EndIf
+
 	Return SetError(__WD_Error($sFuncName, 0, $sJSON), 0, $sJSON)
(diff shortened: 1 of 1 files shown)


wd_helper.au3
```diff
@@ -3221,6 +3221,12 @@ Func _WD_JsonActionKey($sType, $sKey, $iSuffix = Default)
 	Json_Put($vData, '.actions[0].type', $sType)
 	Json_Put($vData, '.actions[0].value', $sKey)
 	Local $sJSON = Json_Encode($vData)
+
+	; Don't encode backslash of Unicode character
+	If StringLeft($sKey,2) = '\u' Then
```
mlipok, line 3226: hmm.... why not :
```
If StringLeft($sKey,2) = '\\u' Then
```
?

Danp2: `$sKey` is the incoming parameter from the calling routine, so the expectation is that this will be in the correct format (ie: \ue009). The proposed change is to fix the situation where the call to Json_Encode escapes the backslash by adding a second one. Therefore, I believe that the proposed change better addresses the situation.

PR #534, opened 2025-07-15
wd_demo.au3 - output in @compiled mode

mlipok:
setting the default value for the compiled version to be able to see the logs

### Proposed changes

in compiled version you should see logs thus default output to console is not good option

### Checklist

- [X] I have read and noticed the [CODE OF CONDUCT](https://github.com/Danp2/au3WebDriver/blob/master/docs/CODE_OF_CONDUCT.md) document
- [X] I have read and noticed the [CONTRIBUTING](https://github.com/Danp2/au3WebDriver/blob/master/docs/CONTRIBUTING.md) document

### Types of changes

- [ ] Other (please describe)

output is set to `ConsoleWrite`

### What is the new behavior?

output is set to `_DebugOut`

none

### Additional context

none

### System under test

not related

### DIFF
--- a/wd_demo.au3
+++ b/wd_demo.au3
@@ -114,6 +114,7 @@ Func _WD_Demo()
 	GUICtrlCreateLabel("ConsoleOut", 15, $iPos + 2)
 	Local $idOutput = GUICtrlCreateCombo("ConsoleWrite", 75, $iPos, 100, 20, $CBS_DROPDOWNLIST)
 	GUICtrlSetData($idOutput, "WD_Console.log|_DebugOut|Null", "ConsoleWrite")
+	If @Compiled Then GUICtrlSetData($idOutput, "_DebugOut") ; set the default value for the compiled version to be able to see the logs
 	#EndRegion - Output
 
 	#Region - demos


Danp2:
While this was never intended to be run as a compiled app, I supposed this change won't cause any issues. 🤔

mlipok:
Recently, I started using "wd_demo.au3" to test problems I encounter on my clients' computers, which I access remotely via TeamViewer.

Especially given the problems
- using WebDriver in #requiredadmin mode
- using WebDriver in conjunction with ESET antivirus software

I think that verifying the correct operation of basic functionality using "wd_demo.exe" is a basic check, the so-called first step when something starts to malfunction.

Because of this, using "wd_demo.exe" became a bit problematic as it was not intended to be run in compiled mode, although it should have been.

So you were right to notice:

> this was never intended to be run as a compiled app

But I think that testing the functionality of "wd_demo" should not be limited to using only its uncompiled version.

For this reason, I proposed this minor functionality change.

It doesn't change the overall functionality of the code, but only makes it easier to work with when using the compiled version.

Since the intended use of the code is usually to compile and run it in its compiled form, it seems reasonable to test not only "wd_demo.au3" but also "wd_demo.exe."

I also believe this because we often use this script as a basic recommendation for users to check if "wd_demo" works.

sven-seyfert:
I strongly believe this change will not cause issues @Danp2 , your assumption is right.
Like @mlipok mentioned, it could be helpful for use cases like he described.

PR #540, opened 2025-08-27
Update wd_demo.au3 - __WD_ConsoleWrite() usage - suplement for - output in @compiled mode

mlipok:
### Proposed changes
Recently I proposed the already merged : https://github.com/Danp2/au3WebDriver/pull/534
Here I propose to continue this topic.

p.s.
As a next step (PR), I'd like to propose using Au3Stripper to handle line numbers in compile mode.

### Checklist

- [ x I have read and noticed the [CONTRIBUTING](https://github.com/Danp2/au3WebDriver/blob/master/docs/CONTRIBUTING.md) document

### Types of changes

- [ ] Other (please describe)

In compiled mode not all testing informotion are outputed to the selected log output.

### What is the new behavior?

In compiled mode all testing informotion are outputed to the selected log output.

None

### Additional context

https://github.com/Danp2/au3WebDriver/pull/534

### System under test

not releated

### DIFF
--- a/wd_demo.au3
+++ b/wd_demo.au3
@@ -196,11 +196,11 @@ Func RunDemo($idDebugging, $idBrowsers, $idUpdate, $idHeadless, $idOutput)
 	Local $sCapabilities = Call($aBrowsers[_GUICtrlComboBox_GetCurSel($idBrowsers)][1], $bHeadless)
 	If @error Then Return SetError(@error, @extended, 0)
 
-	ConsoleWrite("> wd_demo.au3: _WD_Startup" & @CRLF)
+	__WD_ConsoleWrite("> wd_demo.au3: _WD_Startup" & @CRLF)
 	Local $iWebDriver_PID = _WD_Startup()
 	If _RunDemo_ErrorHander((@error <> $_WD_ERROR_Success), @error, @extended, $iWebDriver_PID, $sSession) Then Return
 
-	ConsoleWrite("> wd_demo.au3: _WD_CreateSession" & @CRLF)
+	__WD_ConsoleWrite("> wd_demo.au3: _WD_CreateSession" & @CRLF)
 	$sSession = _WD_CreateSession($sCapabilities)
 	If _RunDemo_ErrorHander((@error <> $_WD_ERROR_Success), @error, @extended, $iWebDriver_PID, $sSession) Then Return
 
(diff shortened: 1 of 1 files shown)


sven-seyfert:
See my commment [here](https://github.com/Danp2/au3WebDriver/pull/541#issuecomment-3236969652). Also relevant for this proposal.

sven-seyfert:
@mlipok, I have to agree with @Danp2.

On a second look I do not understand what do you exactly mean by

> In compiled mode all testing information are outputed to the selected log output.

... which "selected log output"? I am a bit confused to be honest. I reviewed the function:

``` autoit
Func __WD_ConsoleWrite($sMsg, $iDebugLevel = Default, $iError = @error, $iExtended = @extended)
	If $iDebugLevel = Default Or $_WD_DEBUG >= $iDebugLevel Then
		If IsFunc($_WD_CONSOLE) Then
			Call($_WD_CONSOLE, $sMsg & $_WD_CONSOLE_Suffix)
		ElseIf $_WD_CONSOLE = Null Then
			; do nothing
		Else
			FileWrite($_WD_CONSOLE, $sMsg & $_WD_CONSOLE_Suffix)
		EndIf
	EndIf
	Return SetError($iError, $iExtended)
EndFunc   ;==>__WD_ConsoleWrite
```

Are you referring to the `$iDebugLevel`?
A simple `Call(ConsoleWrite, '...' & @CRLF)` each time would do the job, am I right 🤔 ?

Best regards to both of you,
Sven

Danp2:
@sven-seyfert I understand what @mlipok wants. However, I don't agree with this change for the reason mentioned above. There are probably a couple of different ways this could be handled, but I'm not sure that it is worth the effort.

mlipok:
> @mlipok, I have to agree with @Danp2.
> 
> On a second look I do not understand what do you exactly mean by
> 
> > In compiled mode all testing information are outputed to the selected log output.
> 
> ... which "selected log output"? I am a bit confused to be honest. I reviewed the function:


This code is related to the output:

https://github.com/Danp2/au3WebDriver/blob/1834e95206bd4a6ef6952c47a1f1192042f98c0b/wd_demo.au3#L112-L118

You can select output for logs.

When you test PC machine if all is fine with WebDriver and with this UDF, for the first test, you choose the simplest possible solution, which should work always and everywhere.
The simplest example of this in my opinion is WD_DEMO.

Therefore, when you use WD_DEMO to test how AutoIt works in this UDF, regardless of whether you test it in the uncompiled or compiled version, you should receive the same messages in the output.

This is my approach after my recent experiences with startup problems (as ADMIN and in an environment with some Antiviruses).

The proposed changes do not alter the way WebDriver is used per se.

They are intended solely to standardize the display of diagnostic messages in both COMPILED and UNCOMPILATED modes.

sven-seyfert:
Understood, thank you @mlipok.

Danp2:
> The simplest example of this in my opinion is WD_DEMO.

I think there are some that would disagree with this statement. 😆 

> They are intended solely to standardize the display of diagnostic messages in both COMPILED and UNCOMPILATED modes.

The output is already consistent IMO. It just doesn't perform the way that you would prefer. 😉 

> __WD_ConsoleWrite is intended to be an internal function, so using it in this manner isn't setting a good example IMO.

I still stand by this. Have you considered any other alternatives to achieve the desired outcome?

PR #546, opened 2026-04-17
Fix _WD_UpdateDriver to handle changes in msedgedriver updates.

wdouglascampbell:
### Proposed changes

Currently, it does not work to update msedgedriver to the latest version because the update method tries to determine the latest available version which does not work in the case of msedgedriver any longer.  Therefore, instead we want to get the current browser version and use that version to upgrade msedgedriver when our current version doesn't match.  We want to leave the current method working for other browsers.

### Checklist

### Types of changes

- [ ] Other (please describe)

My tool was failing to run because the _WD_UpdateDriver() function was failing to update msedgedriver.

### What is the new behavior?

Now, when I call _WD_UpdateDriver(), it properly evaluates and updates msedgedriver to the version matching my browser.

There should be no adverse effects for other browsers since we check which browser we are working with and if it is not edge, the existing methods of finding the latest release are used.  It is only if the browser we are working with is edge that the new method will be used.

### Additional context

None.

### System under test

- OS: Windows 11 25H2
- OS Arch.: X64
- Browser Edge
- Browser version 147.0.3912.72


Danp2 (commented):
Thanks for the submission. However, I don't believe this will work in all cases as there isn't a guarantee that there will always be a version of msedgedriver that exactly matches the current MSEdge version.

FWIW, one option that I have under consideration is to parse the contents of https://msedgedriver.microsoft.com/listing.json to obtain the needed information.

wdouglascampbell:
Hey Dan, I think you are correct.  I had read somewhere that Microsoft had switched how they were deploying this but now after further research I am thinking that may have been an AI hallucination :(  Or perhaps I am hallucinating now :)

The good news is that I think I found that just like the PR that I submitted last year for the download URL, I think the solution is just changing the URL for this.  I will confirm tomorrow and submit another PR with a fix if that is the case which will be a much, much better solution.

Thanks!

